PR #46, opened 2019-06-25
Change in defualts for stack and aux.xml creation

ehavazli:
stack generation True by default
aux.xml file creation suppressed by switching GDAL_PAM_ENABLED between 'NO' and 'YES' before and after extraction of metadata

### DIFF
--- a/tools/bin/ariaTSsetup.py
+++ b/tools/bin/ariaTSsetup.py
@@ -12,6 +12,7 @@ import glob
 import numpy as np
 from datetime import datetime, time
 from osgeo import gdal
+
 gdal.UseExceptions()
 #Suppress warnings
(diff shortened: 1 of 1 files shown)


tools/bin/ariaTSsetup.py
```diff
@@ -94,6 +96,7 @@ def generateStack(aria_prod,inputFiles,outputFileName,workdir='./'):
     incAng = extractMetaDict(aria_prod,'incidenceAngle')
     lookAng = extractMetaDict(aria_prod,'lookAngle')
     azimuthAng = extractMetaDict(aria_prod,'azimuthAngle')
+    os.environ['GDAL_PAM_ENABLED'] = 'YES'
```
dbekaert: looks good!

PR #55, opened 2019-07-02
hardcode overlap stitching; uniform spelling

bbuzz31:


sssangha:
Just to clarify, why are we hardcoding overlap stitching?

dbekaert:
For now its our only working variant for stitching of the unwrapped layer, so do not want the "2stage" as option for users to try out. Internal we will keep it so we can develop and test the "2stage" implementation. Once matured, we will add it back in as option to use from the command line parser.

PR #57, opened 2019-07-03
Add option to specify output format in tsSetup.py.

sssangha:


dbekaert:
Thanks for confirming @sssangha. Will merge and close.

PR #60, opened 2019-07-09
Combined existing GSHHS-based mask with average amplitude and osm data.

sssangha:
Inspired by Brett's code, I combined the existing GSHHS-based mask with osm and average amplitude data.

There are a couple of caveats though:
1. Computing the average coherence does seem rather slow.
2. The default amplitude threshold (<500) may be a bit liberal for certain cases. E.g. for the San Francisco bay test region, I proceed with only one IFG and found a lot of land masked out inland along most ridges. There is a flag that allows the user to adjust this amplitude threshold, but as we talked about last week some more trial-and-error is necessary across a wider range of study areas to settle on the best default threshold.

test_ampmask

--Sim


sssangha:
Ok, I added in the None option. Plus I also made sure now that all plots + the average coherence raster are stored in a separate "figures" subdirectory. Also, in case the user wishes to redo the average coherence, the program now checks for its existence first and deletes it. Otherwise, the program would append to the existing file and we'd end up with something that isn't an average raster!

tools/ARIAtools/extractProduct.py
```diff
@@ -123,13 +124,17 @@ def prep_dem(demfilename, bbox_file, prods_TOTbbox, proj, arrshape=None, workdir
 
     return demfilename, demfile, Latitude, Longitude
 
-def prep_mask(maskfilename, bbox_file, prods_TOTbbox, proj, arrshape=None, workdir='./', outputFormat='ENVI'):
+def prep_mask(product_dict, maskfilename, bbox_file, prods_TOTbbox, proj, amp_thresh='500', arrshape=None, workdir='./', outputFormat='ENVI'):
```
dbekaert: default to None

tools/ARIAtools/productPlot.py
```diff
@@ -32,6 +32,7 @@ def createParser():
     parser.add_argument('-w', '--workdir', dest='workdir', default='./', help='Specify directory to deposit all outputs. Default is local directory where script is launched.')
     parser.add_argument('-b', '--bbox', dest='bbox', type=str, default=None, help="Provide either valid shapefile or Lat/Lon Bounding SNWE. -- Example : '19 20 -99.5 -98.5'")
     parser.add_argument('-m', '--mask', dest='mask', type=str, default=None, help="Path to mask file or 'Download'. File needs to be GDAL compatabile, contain spatial reference information, and have invalid/valid data represented by 0/1, respectively. If 'Download', will use GSHHS water mask")
+    parser.add_argument('-at', '--amp_thresh', dest='amp_thresh', default=500, type=str, help='Amplitude threshold below which to mask. Specify "None" to not use amplitude mask.')
```
dbekaert: default to None

tools/ARIAtools/tsSetup.py
```diff
@@ -36,6 +36,7 @@ def createParser():
     parser.add_argument('-p', '--projection', dest='projection', default='WGS84', type=str, help='projection for DEM. By default WGS84.')
     parser.add_argument('-b', '--bbox', dest='bbox', type=str, default=None, help="Provide either valid shapefile or Lat/Lon Bounding SNWE. -- Example : '19 20 -99.5 -98.5'")
     parser.add_argument('-m', '--mask', dest='mask', type=str, default=None, help="Path to mask file or 'Download'. File needs to be GDAL compatabile, contain spatial reference information, and have invalid/valid data represented by 0/1, respectively. If 'Download', will use GSHHS water mask")
+    parser.add_argument('-at', '--amp_thresh', dest='amp_thresh', default=500, type=str, help='Amplitude threshold below which to mask. Specify "None" to not use amplitude mask.')
```
dbekaert: set to None by default

PR #65, opened 2019-07-16
New tropo_corrections function to apply GACOS corrections

sssangha:
To use new function, use "-tp" flag to pass path to untared directory containing extracted GACOS product or a list of tarfile paths to extract. The code can handle cases where GACOS products first need to be stitched--stitched GACOS products are exported to the "merged_GACOS" subfolder created inside the specified work directory. 

The program also reports to the user if there is any spatiotemporal discrepancy between the downloaded GACOS product and the extracted phase. 

A reference point for each IFG is estimated following the method outlined in the original GACOS code, e.g. the pixel closest to the nanmean of the phase-field.

Also passed a couple of miscellaneous updates: (1) in ARIAProduct only a comma-delimited input list of files is now allowed as opposed to a space-delimited list (to reflect the documentation). (2) Minor updates to the function description in the "export_products" function.

--Sim

### DIFF
--- a/tools/ARIAtools/ARIAProduct.py
+++ b/tools/ARIAtools/ARIAProduct.py
@@ -43,8 +43,8 @@ class ARIA_standardproduct: #Input file(s) and bbox as either list or physical s
 
         ### Determine if file input is single file, a list, or wildcard.
         # If list of files
-        if len([str(val) for val in filearg.split()])>1:
-            self.files=[str(val) for val in filearg.split()]
+        if len([str(val) for val in filearg.split(',')])>1:
+            self.files=[str(val) for val in filearg.split(',')]
         # If single file or wildcard
         else:
             # If single file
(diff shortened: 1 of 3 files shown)


dbekaert:
@pymonger any idea why circle ci would fail as we have not implemented tests. Is there a change we should be doing when issuing PR?

@sssangha You think it would be good to make a tropo stand-alone program. This will change in future with our ray-tracing tools and want to avoid hooking it in too deeply into extract product incase we decide to make changes?

sssangha:
@dbekaert yeah that was/still is the plan. Even that aside though, clearly the function itself is currently huge and could be split into several different functions or translated into a class in its own right.

Before I work on that, I just wanted some input on the current structure I have outlined before I go too far off on a particular direction over the other. May some of you may have some suggestions on what is the best way to organize this functionality (e.g. should we make it into a class like in ARIAProduct.py).

PR #66, opened 2019-07-17
New tropo_corrections function to apply GACOS corrections

sssangha:
*NOTE, disregard previous branch concerning these changes! 

To use new function, use "-tp" flag in ariaExtract.py or ariaTSsetup.py to pass path to untared directory containing extracted GACOS product or a list of tarfile paths to extract. The code can handle cases where GACOS products first need to be stitched--stitched GACOS products are exported to the "merged_GACOS" subfolder created inside the specified work directory. 

The program also reports to the user if there is any spatiotemporal discrepancy between the downloaded GACOS product and the extracted phase. 

A reference point for each IFG is estimated following the method outlined in the original GACOS code, e.g. the pixel closest to the nanmean of the phase-field. 

Also passed a couple of miscellaneous updates: (1) in ARIAProduct only a comma-delimited input list of files is now allowed as opposed to a space-delimited list (to reflect the documentation). (2) Minor updates to the function description in the "export_products" function.

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -126,7 +126,7 @@ def prep_dem(demfilename, bbox_file, prods_TOTbbox, proj, arrshape=None, workdir
 
     return demfilename, demfile, Latitude, Longitude
 
-def prep_mask(product_dict, maskfilename, bbox_file, prods_TOTbbox, proj, amp_thresh='500', arrshape=None, workdir='./', outputFormat='ENVI'):
+def prep_mask(product_dict, maskfilename, bbox_file, prods_TOTbbox, proj, amp_thresh=None, arrshape=None, workdir='./', outputFormat='ENVI'):
     '''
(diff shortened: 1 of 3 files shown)


dbekaert:
@pymonger any idea why circle CI fails even when we do not have tests?

tools/ARIAtools/extractProduct.py
```diff
@@ -417,7 +416,7 @@ def finalize_metadata(outname, bbox_bounds, prods_TOTbbox, dem, lat, lon, mask=N
                 out_interpolated[iz, iline, ipix] = interp_2d(line, pixel, hgt)
     out_interpolated=np.flip(out_interpolated, axis=0)
     # must use 'nearest' interpolation to avoid disprepancies between different crop extents
```
dbekaert: update this comment as this is no longer accurate.

sssangha: Ok, I fixed that.

PR #69, opened 2019-07-23
prep_mask no longer missing first argument in productPlot

bbuzzanga:


sssangha (commented):
I actually simultaneously fixed this on my own end for the commit I'm gearing up to push.

We need to just pass the entries relevant for the amplitude layers, not the whole dictionary. E.g. "[[j['amplitude'] for j in standardproduct_info.products[1]],"

dbekaert:
@sssangha can we close this as you already did it?

dbekaert:
already accounted by @sssangha

PR #72, opened 2019-07-24
Multithreading now supported. Other changes.

sssangha:
Multithreading now supported. Average rasters now computed directly and efficiently through gdal warp. Multiple other miscellaneous changes:

1. Multithreading now supported for gdalwarp calls in "extractProduct.py" "tsSetup.py" and "productPlot.py". The -nt flag can be used specify number of threads used (by default 2, but the user can also specify "all" to use all available threads). 
2. Masks and DEMs now saved to separate subdirectories, and uncropped/cropped versions of each are saved. If a user launches another run later in the same workspace and doesn't use the cropped mask/DEM that had already been generated, the existing mask/DEM subdirectories are emptied.
3. Removed redundant specification of destination/source no-data value when downloading the DEM in the "prep_dem" function.
4. Now in the "prep_mask" function a user specifies a amplitude threshold to generate a water mask, the average amplitude raster is kept instead of being deleted.
5. Proposed update to Brett's request to pass amplitude layers from dictionary in "prep_mask" call in "productPlot.py".
6. Originally in "tsSetup.py", the single LOS grid generated corresponding to the first product in the stack may not encompass the entire extent of the track in the case the user specifies the "croptounion" option. Hence, an average LOS grid using all files is now passed. 
7. Finally computation of average coherence raster is simplified as averaging is handled directly through gdalwarp, not iteratively through a loop (which would produce errouneous mean estimates for "croptounion" cases).

### DIFF
--- a/tools/ARIAtools/ARIAProduct.py
+++ b/tools/ARIAtools/ARIAProduct.py
@@ -43,8 +43,8 @@ class ARIA_standardproduct: #Input file(s) and bbox as either list or physical s
 
         ### Determine if file input is single file, a list, or wildcard.
         # If list of files
-        if len([str(val) for val in filearg.split()])>1:
-            self.files=[str(val) for val in filearg.split()]
+        if len([str(val) for val in filearg.split(',')])>1:
+            self.files=[str(val) for val in filearg.split(',')]
         # If single file or wildcard
         else:
             # If single file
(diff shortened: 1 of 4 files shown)


sssangha:
Also forgot to mention this commit fixes a potential layer input issue. If there was a trailing space for a given layername in a list (e.g. 'bperp, amplitude'), then the program would have originally exited out with an error trying to find the nonexistant key " amplitude". Now such trailing spaces are removed.

dbekaert:
@sssangha should this not be issues to the dev branch?

dbekaert:
@sssangha thanks for all these changes.

Few changes after looking at it.
- can you resolve the conflicts. It looks like the tropo is considered a new commit, which makes me suspect that you are behind the main branch.
- Check what the default multi-core option is for GDAL. We should remain in sync with when the option is not specified.

PR #76, opened 2019-07-25
Fixed minor bug in argparse for num_thread default.

sssangha:
Default number of threads should be set to string '2', not int 2.


dbekaert (approved):

dbekaert:
looks good.

PR #79, opened 2019-07-27
Bug fixes in prep_dem and prep_mask functions

sssangha:
1. Downloaded DEM not cropped to product/bbox cutline
2. Fixed check in mask/DEM that had erroneously overwritten existing mask/DEM even if the user wished to keep them.


dbekaert (approved):
Looks good, we should also backport to v1 branch

sssangha:
Yup, just pushed that in to v1 too

PR #80, opened 2019-07-27
Backport recent dev commit #d0c33e5 to v1 branch

sssangha:
1. Downloaded DEM not cropped to product/bbox cutline
 2. Fixed check in mask/DEM that had erroneously overwritten existing mask/DEM even if the user wished to keep them.


dbekaert (approved):
looks good

PR #82, opened 2019-07-30
DEM/mask bug-fixes/enhancements. Duplicate products now removed

sssangha:
Bug-fixes/enchancements:
1. Fixed bug in prep_dem/prep_mask which may lead to cropping inconsistent with interferometric grid when user provides custon DEM/mask.
2. In continuous_time, search for duplicate standard products, and then print warning and delete duplicate products from list.
3. Clean up sections where %overlap is computed for interferometric grid vs user bounding box in merged_productbbox

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -593,7 +593,7 @@ def tropo_correction(full_product_dict, tropo_products, bbox_file, prods_TOTbbox
         save_shapefile(i+'.shp', bbox, 'GeoJSON')
         per_overlap=((user_bbox.intersection(open_shapefile(i+'.shp', 0, 0)).area)/(user_bbox.area))*100
         if per_overlap!=100. and per_overlap!=0.:
-            print("WARNING: Common track extent only has %d%% overlap with bbox"%per_overlap+'\n')
+            print("WARNING: Common track extent only has %d%% overlap with tropospheric product %s"%(per_overlap, i)+'\n')
         if per_overlap==0.:
             raise Exception('No spatial overlap between tropospheric product %s and defined bounding box. Resolve conflict and relaunch'%(i))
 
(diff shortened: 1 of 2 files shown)


tools/ARIAtools/extractProduct.py
```diff
@@ -596,11 +591,7 @@ def tropo_correction(full_product_dict, tropo_products, bbox_file, prods_TOTbbox
         bbox=Polygon(np.column_stack((np.array([bbox[2],bbox[3],bbox[3],bbox[2],bbox[2]]),
                             np.array([bbox[0],bbox[0],bbox[1],bbox[1],bbox[0]]))))
         save_shapefile(i+'.shp', bbox, 'GeoJSON')
-        bbox_area=open_shapefile(i+'.shp', 0, 0)
-        bbox_area=user_bbox.intersection(bbox_area)
-        bbox_area=bbox_area.bounds
-        bbox_area=(max(bbox_area[0],bbox_area[2]) - min(bbox_area[0],bbox_area[2]))*(max(bbox_area[1],bbox_area[3]) - min(bbox_area[1],bbox_area[3]))
-        per_overlap=(prods_bbox_area/bbox_area)*100
+        per_overlap=((user_bbox.intersection(open_shapefile(i+'.shp', 0, 0)).area)/(user_bbox.area))*100
```
dbekaert, line 594: Should the message below not be related to overlap warning for tropo?
Its the same message as for a regular GUNW product, so might be confusing.

sssangha: Yeah, true. How about something like this: "print("WARNING: Common track extent only has %d%% overlap with tropospheric product %s"%(per_overlap, i)+'\n')"

Where "i" points to the tropospheric product referenced in the loop?

dbekaert: yes that works.

sssangha: Ok, committed.

dbekaert (approved):

PR #84, opened 2019-07-30
New tool to create bounding box files

ehavazli:
- small script to convert google earth polygons (saved as kml file) to geojson files.


dbekaert (commented):
Looks good, perhaps someone else could do a test.
@ehavazli Would this add new requirements?

ehavazli:
> @ehavazli Would this add new requirements?

No, only gdal is required

dbekaert (approved):

PR #85, opened 2019-07-31
tsSetup bug for when orbit direction is captured.

sssangha:
On line 156, must split the actual file name for '-', as you run into issues when a user specifies a path with -'s.


dbekaert (approved):
Can you also backport?

PR #87, opened 2019-07-31
bug with duplicate product capture.

sssangha:
Backport with v1.

Looks like a bug-fix slipped through the cracks. This patches up the case where a user were to only specify two products which happened to be duplicates.


dbekaert:
@sssangha if its a backport need to commit to v1.

sssangha:
@dbekaert actually what I meant to say that this particular issue was fixed in v1, but not in dev.

dbekaert (approved):

PR #88, opened 2019-08-01
Fixed bug in prep_dem and prep_mask.

sssangha:
1. Fixed bug in prep_dem and prep_mask where the check for existing dem/mask files fails if a user specified an absolute path.
2. If user wishes to perform tropo corrections, the code now checks for the bare minimum files (unw phase and los files) to be produced, even if the user specifies no layers.


dbekaert:
Looks good, can you backport item 1 to the v1 branch?

dbekaert (approved):

PR #89, opened 2019-08-01
Backport: Fixed bug in prep_dem and prep_mask.

sssangha:
Fixed bug in prep_dem and prep_mask where the check for existing dem/mask files fails if a user specified an absolute path.


dbekaert:
@sssangha the codacy report (see under the checks of the PR) flags a few functions imports are un-used. Can you have a look as well.

sssangha:
@dbekaert ok done!

dbekaert (approved):
Looks good, if we do not need them we should also check in dev if we needed them or not.
 Can be done in a next commit.

PR #90, opened 2019-08-01
Fix bug with tropo product shapefile generation.

sssangha:
1. Fixed bug where coordinates were erroneously flipped when generating shapefiles for tropo products.
2. Removed two function calls that are no longer used.


dbekaert (approved):
looks good

PR #95, opened 2019-08-02
add environment.yml for conda create env

yunjunz:
This PR 

+ adds environment.yml file to config the ARIA-tools conda environment, with pip install h5py, so that all dependencies of ARIA-tools and mintpy/prep_aria.py can be installed in one line command as the `ARIA-tools` environment.

+ updates README.md to use `conda env create` instead of `conda create -n`


dbekaert (approved):
looks good. Thanks for the contribution @yunjunz

PR #96, opened 2019-08-06
Fix bug which unnecessarily deleted existing files.

sssangha:
In the productPlot script, existing "avgcoherence*'" files were deleted before computing average coherence rasters. This is not necessary as existing files would be overwritten, plus this would have the unintended consequence of deleting other plots the user may wish to generate in the same command-line input. E.g. if the user ran "ariaPlot.py -f 'products/S1-*' -verbose -w test_plotseparate -nt 8 --plotcoh --plotbperpcoh --makeavgoh", the requested coherence plot will be deleted by this check.

### DIFF
--- a/tools/ARIAtools/productPlot.py
+++ b/tools/ARIAtools/productPlot.py
@@ -366,13 +366,7 @@ class plot_class:
         '''
             Generate average coherence raster.
         '''
-
-        # Import functions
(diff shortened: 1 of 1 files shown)


dbekaert (approved):
Thnx for fixing this. Can you also backport?

PR #103, opened 2019-08-08
Suppress matplotlib warnings

sssangha:
Transparency warnings now suppressed in productPlot


dbekaert:
believe we merged this already so closing it.

PR #107, opened 2019-08-22
add function to plot the temporally averaged spatial coherence

bbuzzanga:


sssangha:
@bbuzz31 @dbekaert  ok, I looked over the code and it looks great! I just made some relatively minor edits to the new plotting functionality.

tools/ARIAtools/extractProduct.py
```diff
@@ -645,7 +645,7 @@ def tropo_correction(full_product_dict, tropo_products, bbox_file, prods_TOTbbox
             if os.path.exists(os.path.join(outDir,'lookAngle',product_dict[2][i][0])):
                 lookfile=gdal.Open(os.path.join(outDir,'lookAngle',product_dict[2][i][0])).ReadAsArray()
             else:
-                lookfile=gdal.Open(os.path.join(outDir,'lookAngle',product_dict[2][i][0])).ReadAsArray()
+                lookfile=gdal.Open(os.path.join(outDir,'lookAngle',product_dict[2][0][0])).ReadAsArray()
```
dbekaert, line 648: @sssangha this does not cover the case when the first product does not cover it completely. Should this not be in line with what we do for time-series?

dbekaert (approved):
spoke with Sim and confirm the issue i flagged, can be closed.

PR #108, opened 2019-08-26
Pass tropo-corrected layers to stack VRT

sssangha:
If tropo corrections specified, corrected phase layers passed to stack VRT.

Before, original unw phase layers where being passed, even if user requested tropo corrections.


dbekaert (approved):

dbekaert:
might in future come up with a better naming convention for the troposphere when we add in the RAiDER package.

PR #110, opened 2019-08-27
fix bug in VRT naming

bbuzzanga:


sssangha:
@bbuzz31 @dbekaert looks good to me!

sssangha:
@bbuzz31 @dbekaert on second though, give me a few minutes! Tracked down another minor bug. Testing now, will update accordingly

bbuzzanga:
Would also be a good idea to by default add all products to the stack, regardless if they have tropo corrections
Currently, it will only add tropo if they are requested

sssangha:
@bbuzz31 I'm not sure about mixing corrected and uncorrected products. If the user explicitly wishes to perform atmospheric corrections, those layers should remain segregated from the uncorrected ones.

sssangha:
@dbekaert @bbuzz31 ok, fixed the other bug I found!

dbekaert:
> @bbuzz31 I'm not sure about mixing corrected and uncorrected products. If the user explicitly wishes to perform atmospheric corrections, those layers should remain segregated from the uncorrected ones.

We should not really be mixing products.
Either they are tropo corrected or they are not.
Should be explicit request from user.

tools/ARIAtools/tsSetup.py
```diff
@@ -114,7 +114,7 @@ def generateStack(aria_prod,inputFiles,outputFileName,workdir='./'):
         print('Number of unwrapped interferograms discovered: ', len(intList))
         Dlist = intList
     elif inputFiles in ['tropocorrected_products', 'tropocorrected', 'tropo']:
-        domainName = 'tropocorrected_products'
+        domainName = 'unwrappedPhase'
```
dbekaert, line 117: Ok, i understand this is needed given mintpy is not aware of us making any corrections.
Looks good.

dbekaert (approved):

PR #117, opened 2019-09-19
Change no data value from 0 -> 2

bbuzzanga:
no data value of 0 in watermask which is either 0/1 causes problem when loaded into GMT


dbekaert:
@brett mintpy does not always track no-data values, this might have implications for TS processing.

@ehavazli @sssangha Can you comment what the implications are for this in minpy.

sssangha:
@bbuzz31 We should keep the mask binary as it was originally. This is because the rasters are directly multiplied by the masks before output. An alternative to multiplication is to mask the rasters along the lines of something as so: raster[raster == masknodatavalue] = rasternodatavalue

ehavazli:
@bbuzz31 @sssangha @dbekaert 
I agree with Sim on keeping the mask binary. Other than MintPy it may have implications when working with different software in the future. A GMT based solution is the way to go on this one.

bbuzzanga:
Yes with the multiplication a non-binary solution seems inappropriate. 
However I think there will be unexpected behavior for software other than GMT which will similarly set water pixels to nan.

EJFielding:
I import many datasets that have no-data values of 0 in GMT and QGIS. It might require an extra grdmath step in GMT to convert the 0 pixels to no-data (NaN), but usually one can apply the mask using a grdmath IF or EQ operation.

dbekaert:
Lets keep the original, we track the no-data properly into the meta-data. See it as user responsibility to do it in their preferred tools.

PR #119, opened 2019-09-26
Downsampling routines and dropping of IFGs not meeting spatial percentage requirement

rzinke:
Hello all, 

I have incorporated a set of optional resampling routines in case the products need to be downsampled for timeseries creation. The modified scripts are tsSetup and extractProducts. 

I have tested these implementations on ARIA Track 41 data -- using the full dataset for the non-tropo-corrected products; and using a limited dataset for the tropo-corrected products. 

Could you please review these changes when you get the chance and let me know if these work for your system? 

Cheers,
Rob

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -43,6 +43,7 @@ def createParser():
     parser.add_argument('-of', '--outputFormat', dest='outputFormat', type=str, default='VRT', help='GDAL compatible output format (e.g., "ENVI", "GTiff"). By default files are generated virtually except for "bPerpendicular", "bParallel", "incidenceAngle", "lookAngle","azimuthAngle", "unwrappedPhase" as these are require either DEM intersection or corrections to be applied')
     parser.add_argument('-croptounion', '--croptounion', action='store_true', dest='croptounion', help="If turned on, IFGs cropped to bounds based off of union and bbox (if specified). Program defaults to crop all IFGs to bounds based off of common intersection and bbox (if specified).")
     parser.add_argument('-verbose', '--verbose', action='store_true', dest='verbose', help="Toggle verbose mode on.")
+    parser.add_argument('-r', '--resolution', dest='cell_resolution', default=0.000833333333333, type=float, help='Pixel resolution in degrees. Default = 3 arcsec = 0.000833333333333 degrees.')
 
     return parser
 
@@ -87,7 +88,7 @@ class InterpCube(object):
         est = interp1d(self.hgts, vals, kind='cubic')
         return est(h) + self.offset
 
-def prep_dem(demfilename, bbox_file, prods_TOTbbox, proj, arrshape=None, workdir='./', outputFormat='ENVI', num_threads='2'):
+def prep_dem(demfilename, bbox_file, prods_TOTbbox, proj, arrshape=None, workdir='./', outputFormat='ENVI', num_threads='2', cell_resolution=None):
     '''
         Function to load and export DEM, lat, lon arrays.
         If "Download" flag is specified, DEM will be donwloaded on the fly.
@@ -254,7 +255,7 @@ def prep_mask(product_dict, maskfilename, bbox_file, prods_TOTbbox, proj, amp_th
 
     return mask
 
-def merged_productbbox(product_dict, workdir='./', bbox_file=None, croptounion=False, num_threads='2'):
+def merged_productbbox(product_dict, workdir='./', bbox_file=None, croptounion=False, num_threads='2', cell_resolution=None):
     '''
         Extract/merge productBoundingBox layers for each pair and update dict, report common track bbox (default is to take common intersection, but user may specify union), and expected shape for DEM.
     '''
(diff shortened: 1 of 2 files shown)


tools/ARIAtools/extractProduct.py
```diff
@@ -267,20 +267,48 @@ def merged_productbbox(product_dict, workdir='./', bbox_file=None, croptounion=F
...
+                # find area of intersection between user bbox and frame extent
+                prod_in_user_bbox=prods_bbox.intersection(user_bbox)
+
+                # report percent of frame within user bbox
+                per_in_user_bbox=prod_in_user_bbox.area/prods_bbox.area*100
+
+                # remove frame if area less than some limit
+                if per_in_user_bbox>min_percent_frame_overlap:
+                    frames_in_scene+=1
+
```
dbekaert, line 297: @rzinke From a quick read through the code, i would not do the percentage overlap check in individual products. Rather would do the check on the union of the products for a given InSAR pair. @sssangha to provide some additional discussion.

tools/ARIAtools/extractProduct.py
```diff
@@ -267,20 +267,48 @@ def merged_productbbox(product_dict, workdir='./', bbox_file=None, croptounion=F
...
             if os.path.exists(outname):
                 union_bbox=open_shapefile(outname, 0, 0)
```
dbekaert, line 299: @rzinke @sssangha checking the percentage overlap here.

tools/ARIAtools/extractProduct.py
```diff
@@ -407,11 +453,42 @@ def export_products(full_product_dict, bbox_file, prods_TOTbbox, layers, dem=Non
                     outFileUnw=os.path.join(outDir,'unwrappedPhase',product_dict[1][i[0]][0])
                     outFileConnComp=os.path.join(outDir,'connectedComponents',product_dict[1][i[0]][0])
 
+                    # change outputFormat if resampling is specified
+                    if arrshape is not None and outputFormat=='VRT':
```
dbekaert, line 457: @rzinke There should be a way to keep  VRT even when downsampling. Similar to what we do for the coherence VRT having a suffix "full" for the uncropped verison and the cropped being just the date pairs. We definitely want to avoid the need to generate ENVI file if possible.@sssangha what do you think.

You could also leave the full being the uncropped full resolution version, and have the cropped and the downsampled being the regular date pair combination.

tools/ARIAtools/extractProduct.py
```diff
@@ -407,11 +453,42 @@ def export_products(full_product_dict, bbox_file, prods_TOTbbox, layers, dem=Non
...
+                        outputFormat='ENVI'
+
                     # calling the stitching methods
+                    print('Stitching ({} method)'.format(stitchMethodType))
                     if stitchMethodType == 'overlap':
-                        product_stitch_overlap(unw_files,conn_files,prod_bbox_files,bounds,prods_TOTbbox, outFileUnw=outFileUnw,outFileConnComp= outFileConnComp,mask=mask,outputFormat = outputFormat,verbose=verbose)
+                        product_stitch_overlap(unw_files,conn_files,prod_bbox_files,bounds,prods_TOTbbox, outFileUnw=outFileUnw,outFileConnComp=outFileConnComp,outputFormat = outputFormat,verbose=verbose)
                     elif stitchMethodType == '2stage':
-                        product_stitch_2stage(unw_files,conn_files,bounds,prods_TOTbbox,outFileUnw=outFileUnw,outFileConnComp= outFileConnComp,mask=mask,outputFormat = outputFormat,verbose=verbose)
+                        product_stitch_2stage(unw_files,conn_files,bounds,prods_TOTbbox,outFileUnw=outFileUnw,outFileConnComp= outFileConnComp,outputFormat = outputFormat,verbose=verbose)
```
dbekaert, line 465: @rzinke @sssangha I noticed the mask option has been removed here from the product_stitch_2stage. Did not see an edit to the function itself. Double check if this was not done in error.

dbekaert:
Also double check the codacy requirements. Looks like in the previous commit there was an issue, but for this one its still reviewing.

tools/ARIAtools/extractProduct.py
rzinke: Hi @dbekaert and @sssangha,
Thanks for reviewing the changes. Yes, I remember the original plan was to check the overlap percentage further down in the script, possibly in section below "if bbox_file is not None: ...", but the issue is that sometimes a single frame creeps into the bbox area and a new scene is created because of it. Here is an example of the issue: 

Screen Shot 2019-10-04 at 2 14 02 PM

It seemed like it was easier to throw this one frame out rather than make an entire scene, then delete the entire scene. I also would have to reverse the logic... As implemented above, the script says "the unwanted frame is most outside the user bbox, so throw it out." In the suggested case, I think it would have to say "the unwanted scene does not fill up the user bbox, so throw it out." Is there a more elegant way of doing it? What do you think?

tools/ARIAtools/extractProduct.py
rzinke: Oops. If this was my doing it was certainly not done on purpose.

tools/ARIAtools/extractProduct.py
rzinke: Hi @dbekaert and @sssangha, I have added the suffix "resamp" to the end of the downsampled VRT files to avoid necessarily writing to ENVI format. Overwriting the VRTs produces a RuntimeError that is ignored: 

Exception ignored in: <built-in function delete_Dataset>
RuntimeError: /Users/rzinke/Documents/Tibet/Dsample/Test/connectedComponents/20180719_20180625resamp, band 1: WriteBlock() not supported for this dataset.

A quick search for "WriteBlock() not supported for this dataset" turns up this page: https://stackoverflow.com/questions/10635107/how-to-write-manipulated-raster-values-to-ascii-grid-with-gdal which suggests writing to a different filetype (this error did not appear with the ENVI output format enforced). Seeing as this is not a fatal error, however, and that the routine seems to produce valid timeseries results, perhaps we can ignore it? 

Here is how I have currently rewritten the code: 
                    if arrshape:
                        print('Resampling unwrapped phase: {}'.format(outFileUnw))
                        # resample
                        gdal.Warp(outFileUnw+'resamp', outFileUnw, options=gdal.WarpOptions(format=outputFormat, cutlineDSName=prods_TOTbbox, outputBounds=bounds, width=arrshape[1], height=arrshape[0], resampleAlg='lanczos',multithread=True, options=['NUM_THREADS=%s'%(num_threads)+' -overwrite']))
                        # update VRT
                        gdal.BuildVRT(outFileUnw+'.vrt', outFileUnw+'resamp', options=gdal.BuildVRTOptions(options=['-overwrite']))

PR #120, opened 2019-10-12
Support for downsampling and kicking out short scenes

sssangha:
To allow for downsampling and kick out of short scenes, the following changes were made:
1. Use the "-cr" flag in ariaTSsetup.py and ariaExtract.py to specify a percentage to resample the final outputs wrt the original resolution of a standard product. DEM/mask products are also resampled accordingly, but only after all other layers have already been extracted in order to avoid introducing artifacts when downsampling masked arrays or extracting metadata layers with a downsampled DEM. DEM and mask objects are now tagged with their filenames to reduce complexity when trying to access these names during the resampling stage.
2. The new "resampleRaster" function in vrtManager.py handles the resampling calls.
3. Use the "-po" flag in ariaTSsetup.py and ariaExtract.py to specify a minimum threshold percentage of overlap (by default 5%) products must maintain wrt the specified bounding box. If no bounding box is specified, all products must maintain this threshold instead wrt the longest scene in the stack. References to products that fail to meet this criteria are removed.

Rob, I didn't have a chance to test this out with GACOS corrections. Could you give that a try, in addition to just playing around with the proposed options?

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -433,8 +433,13 @@ def export_products(full_product_dict, bbox_file, prods_TOTbbox, layers, dem=Non
                         product_stitch_overlap(unw_files,conn_files,prod_bbox_files,bounds,prods_TOTbbox, outFileUnw=outFileUnw,outFileConnComp= outFileConnComp, mask=mask,outputFormat = outputFormat,verbose=verbose)
                     elif stitchMethodType == '2stage':
                         product_stitch_2stage(unw_files,conn_files,bounds,prods_TOTbbox,outFileUnw=outFileUnw,outFileConnComp= outFileConnComp, mask=mask,outputFormat = outputFormat,verbose=verbose)
+                        
+                    #If necessary, resample both unw/conn_comp files
+                    if cellResolution is not None:
+                        resampleRaster(outFileUnw, cellResolution, bounds, prods_TOTbbox, outputFormat=outputFormat, num_threads=num_threads)
+                        resampleRaster(outFileConnComp, cellResolution, bounds, prods_TOTbbox, outputFormat=outputFormat, num_threads=num_threads)
             #If necessary, resample raster
-            if cellResolution is not None:
+            if cellResolution is not None and key!='unwrappedPhase' and key!='connectedComponents':
                 resampleRaster(outname, cellResolution, bounds, prods_TOTbbox, outputFormat=outputFormat, num_threads=num_threads)
 
         prog_bar.close()
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -42,6 +42,8 @@ def createParser():
 #    parser.add_argument('-sm', '--stitchMethod', dest='stitchMethodType',  type=str, default='overlap', help="Method applied to stitch the unwrapped data. Either 'overlap', where product overlap is minimized, or '2stage', where minimization is done on connected components, are allowed methods. Default is 'overlap'.")
     parser.add_argument('-of', '--outputFormat', dest='outputFormat', type=str, default='VRT', help='GDAL compatible output format (e.g., "ENVI", "GTiff"). By default files are generated virtually except for "bPerpendicular", "bParallel", "incidenceAngle", "lookAngle","azimuthAngle", "unwrappedPhase" as these are require either DEM intersection or corrections to be applied')
     parser.add_argument('-croptounion', '--croptounion', action='store_true', dest='croptounion', help="If turned on, IFGs cropped to bounds based off of union and bbox (if specified). Program defaults to crop all IFGs to bounds based off of common intersection and bbox (if specified).")
+    parser.add_argument('-cr', '--cellResolution', dest='cellResolution', type=float, default=None, help='Resolution input WRT percentage of original dimensions. E.g. 50. = 50%%"')
+    parser.add_argument('-po', '--percentOverlap', dest='percentOverlap', type=float, default=5., help='Minimum percentage of overlap of scenes wrt specified bounding box. Default 5. = 5%%"')
     parser.add_argument('-verbose', '--verbose', action='store_true', dest='verbose', help="Toggle verbose mode on.")
 
     return parser
@@ -140,6 +142,7 @@ def prep_dem(demfilename, bbox_file, prods_TOTbbox, proj, arrshape=None, workdir
         #pass cropped DEM
         demfile = gdal.Warp('', demfilename, options=gdal.WarpOptions(format="MEM", cutlineDSName=prods_TOTbbox, outputBounds=bounds, dstNodata=0, width=arrshape[1], height=arrshape[0], multithread=True, options=['NUM_THREADS=%s'%(num_threads)]))
         demfile.SetProjection(proj)
+        demfile.SetDescription(demfilename)
 
         ##SS Do we need lon lat if we would be doing gdal reproject using projection and transformation? See our earlier discussions.
         # Define lat/lon arrays for fullres layers
@@ -243,18 +246,16 @@ def prep_mask(product_dict, maskfilename, bbox_file, prods_TOTbbox, proj, amp_th
             mask_file = gdal.Open(maskfilename,gdal.GA_Update)
             mask_file.SetProjection(proj) ; del mask_file
             gdal.Translate(maskfilename+'.vrt', maskfilename, options=gdal.TranslateOptions(format="VRT"))
-
         #pass cropped DEM
         mask=gdal.Warp('', maskfilename, options=gdal.WarpOptions(format="MEM", cutlineDSName=prods_TOTbbox, outputBounds=bounds, width=arrshape[1], height=arrshape[0], multithread=True, options=['NUM_THREADS=%s'%(num_threads)]))
         mask.SetProjection(proj)
-        mask=mask.ReadAsArray()
-
+        mask.SetDescription(maskfilename)
     except:
         raise Exception('Failed to open user mask')
 
     return mask
 
-def merged_productbbox(product_dict, workdir='./', bbox_file=None, croptounion=False, num_threads='2'):
+def merged_productbbox(metadata_dict, product_dict, workdir='./', bbox_file=None, croptounion=False, num_threads='2', percentOverlap=50):
     '''
         Extract/merge productBoundingBox layers for each pair and update dict, report common track bbox (default is to take common intersection, but user may specify union), and expected shape for DEM.
     '''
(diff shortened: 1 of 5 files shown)


dbekaert:
@sssangha not sure if you might have captured earlier discussions with @rzinke.

Can you clarify:
- Does the percentage downsampling imply an integer multiple of the resolution?
- The percentage overlap might not be a great constraint as it does not describe the overlap in square km area. i.e. This is not the same threshold for a small compared to a larger AOI. Perhaps we can convert it to a metric unit threshold. 
- I am unclear why the evaluation needs to be done per product and not per merged product.
e.g. keep all products which covers the bbox, and ensure that the coverage of the merged product meets the requirement.

tools/ARIAtools/extractProduct.py
```diff
@@ -294,22 +300,33 @@ def merged_productbbox(product_dict, workdir='./', bbox_file=None, croptounion=F
...
-        if prods_bbox.bounds==():
-            raise Exception('No common overlap, footprint cannot be generated. Last scene checked: %s'%(scene['productBoundingBox'][0]))
-        save_shapefile(prods_TOTbbox, prods_bbox, 'GeoJSON')
+            # Estimate percentage of overlap with bbox
+            if prods_bbox.bounds==():
+                print('Rejected scene %s has no common overlap with bbox'%(scene['productBoundingBox'][0]))
+                rejected_scenes.append(product_dict.index(scene))
+                os.remove(scene['productBoundingBox'][0])
+            else:
+                per_overlap=(prods_bbox.area/total_bbox.area)*100
```
dbekaert: Originally this is an area, but gets converted to a percentage. An area threshold would be something that can be related across AOI's provided. Threshold based on percentage becomes a bit more AOI specific.

tools/ARIAtools/extractProduct.py
```diff
@@ -294,22 +300,33 @@ def merged_productbbox(product_dict, workdir='./', bbox_file=None, croptounion=F
...
     if bbox_file is not None:
         user_bbox=open_shapefile(bbox_file, 0, 0)
         total_bbox=open_shapefile(prods_TOTbbox, 0, 0)
         user_bbox=user_bbox.intersection(total_bbox)
         save_shapefile(prods_TOTbbox, user_bbox, 'GeoJSON')
-
-        # Estimate percentage of overlap with bbox
-        per_overlap=((user_bbox.area)/(open_shapefile(bbox_file, 0, 0).area))*100
-        if per_overlap<50.:
-            print("WARNING: Common track extent only has %d%% overlap with bbox"%per_overlap+'\n')
```
dbekaert, line 312: At the product stage once could keep it to only products that cover the bbox.
At the merge stage one could compare the threshold of coverage.
The latter is now removed.

tools/ARIAtools/vrtmanager.py
```diff
@@ -66,3 +66,31 @@ def renderOGRVRT(vrt_filename, src_datasets):
...
+
+    # Access original shape
+    ds=gdal.Warp('', fname, options=gdal.WarpOptions(format="MEM", outputBounds=bounds, multithread=True, options=['NUM_THREADS=%s'%(num_threads)]))
+    arrshape=[ds.RasterYSize, ds.RasterXSize] # Get shape of full res layers
+    arrshape=[int(i*(cellResolution/100.)) for i in arrshape]
+
+    # Must resample mask/connected components files with nearest-neighbor
+    if fname.split('/')[-2]=='mask' or fname.split('/')[-2]=='connectedComponents':
+        # Resample raster
+        gdal.Warp(fname, fname, options=gdal.WarpOptions(format=outputFormat, cutlineDSName=prods_TOTbbox, outputBounds=bounds, width=arrshape[1], height=arrshape[0], resampleAlg='near',multithread=True, options=['NUM_THREADS=%s'%(num_threads)+' -overwrite']))
```
dbekaert: Ok nearest for connected component and mask

tools/ARIAtools/extractProduct.py
sssangha: But this percentage is then directly compared to the user-defined percentage of overlap. Are you suggesting then instead to have the user-defined input in units of m^2 or km^2?

tools/ARIAtools/extractProduct.py
sssangha: Yeah, in this function we still compare the overlap of the merged product, not of each individual product. The logic here has been moved up to lines ~305-314

tools/ARIAtools/extractProduct.py
dbekaert: Im referring to a threshold that can be applied across different AOI's and have a physical meaning. The percentage requires knowledge in advance on the AOI extend. E.g. the maximum union etc could change if there is a poor product being generated.

tools/ARIAtools/extractProduct.py
dbekaert: ok thanks for confirming. The comparison on the merged box only is good. Note that that is different from the explanation given in the  PR description.

sssangha:
> @sssangha not sure if you might have captured earlier discussions with @rzinke.
> 
> Can you clarify:
> 
>     * Does the percentage downsampling imply an integer multiple of the resolution?
> 
>     * The percentage overlap might not be a great constraint as it does not describe the overlap in square km area. i.e. This is not the same threshold for a small compared to a larger AOI. Perhaps we can convert it to a metric unit threshold.
> 
>     * I am unclear why the evaluation needs to be done per product and not per merged product.
>       e.g. keep all products which covers the bbox, and ensure that the coverage of the merged product meets the requirement.



> @sssangha not sure if you might have captured earlier discussions with @rzinke.
> 
> Can you clarify:
> 
>     * Does the percentage downsampling imply an integer multiple of the resolution?
> 

@dbekaert Yeah, I just realized the logic of the code as is was too basic as it directly assigned the resampled x/y dimensions (e.g. original raster dimensions * percentage), which doesn't ensure the aspect ratio remains locked. So instead, I am pushing a change that specifies the resolution directly, see the changes in the vrtmanager

tools/ARIAtools/extractProduct.py
rzinke: I agree, if this is practical. I had originally kicked out products based on fraction of the frame area in part for this reason... because the frame areas are relatively consistent, but one doesn't need to know the exact area of a frame in km2. 

But otherwise, I follow your logic that it is better to implement this criterion in the lower part of the script, after the merged area has been defined.

dbekaert: Frames are only consistent within AOI, but not on AOI boudaries.

tools/ARIAtools/vrtmanager.py
```diff
@@ -79,17 +79,16 @@ def resampleRaster(fname, cellResolution, bounds, prods_TOTbbox, outputFormat='E
 
     # Access original shape
     ds=gdal.Warp('', fname, options=gdal.WarpOptions(format="MEM", outputBounds=bounds, multithread=True, options=['NUM_THREADS=%s'%(num_threads)]))
-    arrshape=[ds.RasterYSize, ds.RasterXSize] # Get shape of full res layers
-    arrshape=[int(i*(cellResolution/100.)) for i in arrshape]
+    arrshape=[abs(ds.GetGeoTransform()[1])/(cellResolution/100.), abs(ds.GetGeoTransform()[-1])/(cellResolution/100.)] # Get output res
```
dbekaert: To confirm, this does not imply integer multiple of the resolution.

tools/ARIAtools/extractProduct.py
sssangha: Yeah, kind of straying away from your point for a sec, it seems you would need to have a rough idea of the AOI extent a priori. I've already started to have problems for instance with the CA study area because track 87 exists in both CA and Saudi Arabia.

So I'm not sure of the best way to proceed without at least specifying a rough bbox.

Currently if no bbox is specified, the code takes the longest scene and uses it as a spatial reference (i.e. its area becomes the "bbox"), see lines 289-292.

tools/ARIAtools/vrtmanager.py
sssangha: This isn't necessarily an integer multiple of the natural resolution, but the aspect ratio is preserved, and consistent vis-a-vis the other standard product layers and DEM/mask. 

I'm not sure how to provide more input flexibility than allowing to specify a percentage. E.g. if we force the input to be an integer multiple, I'm not sure of what we can allow for outside of specifying 1-arc,2-arc,4-arc, etc.

sssangha:
@rzinke Ok, check out the latest fix I made to the vrtmanager. Now it handles cases where users just wish to generate VRT files.

rzinke:
Hi @sssangha thanks for all the modifications. I am still getting an error when I try to use the downsampling routine in ariaTSsetup. Here's the error: 

7 out of 30 interferograms rejected since stitched interferogram would have gaps
Thread count specified for gdal multiprocessing = 8
Download/cropping DEM
Downloaded 3 arc-sec SRTM DEM here: TSextract2/DEM/SRTM_3arcsec.dem

Extracting unwrapped phase, coherence, and connected components for each interferogram pair
Generating: unwrappedPhase - [======================= 83% =============>        ] 20181128_20181116  634s /   129sMulti-core version failed, will try single for loop
Generating: unwrappedPhase - [==================================================] 20181222_20181210  758s /    31s
Generating: coherence - [==================================================] 20181222_20181210  262s /    10s

Extracting single incidence angle, look angle and azimuth angle files valid over common interferometric grid
Generating: incidenceAngle - [==================================================] 20180707_20180613
Generating: lookAngle - [==================================================] 20180707_20180613
Generating: azimuthAngle - [==================================================] 20180707_20180613
Creating directory: TSextract2/stack
Number of coherence discovered:  23
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/bin/ariaTSsetup.py", line 15, in <module>
    main(inps)
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.7/site-packages/ARIAtools/tsSetup.py", line 316, in main
    generateStack(standardproduct_info,'coherence','cohStack',workdir=inps.workdir)
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.7/site-packages/ARIAtools/tsSetup.py", line 147, in generateStack
    ds = gdal.Open(data, gdal.GA_ReadOnly)
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.7/site-packages/osgeo/gdal.py", line 3308, in Open 
    return _gdal.Open(*args)
RuntimeError: GDALWarpOptions.Validate(): hSrcDS is not set.

sssangha:
@rzinke @dbekaert listed are the updates I have pushed:

When downsampling (with the --multilooking flag), users now have to specify an integer number instead of a percentage. I.e. multilooking factor is an integer multiple of standard resolution. For example 2 = 90m*2 = 180m. 

Threshold of coverage now controlled by km^2 of overlap, as opposed to percentage of overlap. Equal area projection used.

Bug-fix: If mask specified/provided as input, the connected component file masked values are now assigned as -1 as opposed to 0, in agreement with the connected component nodata values.

Bug-fix: Error resolved when user specifies VRT output format and wishes to downsample outputs.

dbekaert:
@sssangha @rzinke i noticed this PR is still open. Did we get closure on it?

sssangha:
@dbekaert @rzinke 
Refer to my new "rankedresampleVSdefaultresample" branch where I merge the functionality of the two existing branches.

I will leave open the other two aforementioned branches for reference and will delete them once these changes are pushed.

rzinke:
Hi @sssangha, thanks! Could you remind me the difference between ranked resample and default resample? 

I can start testing these this afternoon.

sssangha:
Hello @rzinke, the difference between ranked resample is to determine the mode of the connected component within a given resampling window and only average phase pixels spatially coincident with said mode. In cases where there are multiple modes (e.g. an equal number of the two most frequent connected component pixels), the smallest value is selected.

The default resampling doesn't take into account the connected components and just does a simple average in a given window.

PR #122, opened 2019-10-26
Sss rankedresamplebeta

sssangha:
Building off of commits from the "sss_resampleANDdownselection" branch, downsampling for unw phase is driven by the mode of the connected component pixels over the same corresponding window. E.g., if a user wishes to downsample by a factor of 2, in a given 2x2 connected component window of [[1,2],[2,1]] and corresponding unw phase window of [[3.4,5.0],[5.0,3.4]], the downsampled pixels will be mode(conn_comp)=1 and mean([3.4,3.4])=3.4, respectively.

Currently this pixel function is implemented in a loop, but in the long term it would be more efficient to directly take the mode of the conn_comp file through gdal and code an alternative way to resample the unw phase.

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -435,12 +435,11 @@ def export_products(full_product_dict, bbox_file, prods_TOTbbox, layers, dem=Non
 
                     #If necessary, resample both unw/conn_comp files
                     if multilooking is not None:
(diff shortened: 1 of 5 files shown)


sssangha:
@rzinke I have fixed a bug where the VRT files were improperly updated vis-a-vis the physically extracted layers they referenced, where applicable

PR #123, opened 2019-11-26
fix metadata bugs for tsSetup

yunjunz:
This PR includes the following changes:

1. remote `.data` while asigning the following metadata, otherwise it results in `<memory at >` instead of the real number:
- wavelength
- startRange
- endRange
- rangeSpacing

2. remove unnecessary "> in SRS in the beginging of xml file

3. add line break at the end of <GeoTransform>


dbekaert (approved):
Looks good, thanks for the fix

PR #124, opened 2019-12-18
Multilooking option & spatial area threshold

sssangha:
Building off of commits from the "sss_resampleANDdownselection" and "ss_resampleANDdownselection" branches, downsampling for unw phase is driven by the mode of the connected component pixels over the same corresponding window. E.g., if a user wishes to downsample by a factor of 2, in a given 2x2 connected component window of [[1,2],[2,1]] and corresponding unw phase window of [[3.4,5.0],[5.0,3.4]], the downsampled pixels will be mode(conn_comp)=1 and mean([3.4,3.4])=3.4, respectively. Specify the downsampling factor with the '-ml' flag and pass '-rr' to indicate you wish to use this ranked resampling feature. Otherwise, by default a simple gdal average resampling algorithm is used instead.

Currently this pixel function is implemented in a loop, but in the long term it would be more efficient to directly take the mode of the conn_comp file through gdal and code an alternative way to resample the unw phase.

The user may now specify the minimum overlap threshold with the '-mo' flag. Threshold of coverage now controlled by km^2 of overlap, as opposed to percentage of overlap. Default 0.0081 = 0.0081km2 = area of single pixel at standard 90m resolution. Equal area projection used.

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -435,12 +435,11 @@ def export_products(full_product_dict, bbox_file, prods_TOTbbox, layers, dem=Non
 
                     #If necessary, resample both unw/conn_comp files
                     if multilooking is not None:
-                        resampleRaster(outFileUnw, multilooking, bounds, prods_TOTbbox, outputFormat=outputFormat, num_threads=num_threads)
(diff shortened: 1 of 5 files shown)


dbekaert:
@sssangha Can you fix codacy and then we will merge this.

Issue #129, opened 2020-01-16
Unexpected marker ID when multilooking


rzinke:
Hello, I have been experimenting with the multilooking feature. For my data set, when I set the multilook parameter to 3 and specify a bounding box, the ariaTSsetup.py routine appears to work through all the given products. However, when I try to apply the next step, MintPy's prep_aria.py, I get an error saying one of the files cannot be read. When I try to look at the extracted unwrapped interferogram that is causing the problem using gdalinfo, I get the following error message: 

(ARIA-tools) [rzinke@sar unwrappedPhase]$ gdalinfo 20160903_20160717
ERROR 1: A marker ID was expected (0xff--) instead of 0000cf40

ERROR 1: opj_read_header() failed
gdalinfo failed - unable to open '20160903_20160717'.

This is odd, because all of the other standard products worked in this procedure -- that is, if I toss out that one product, a can run the process from product download to timeseries computation without issue. Furthermore, no errors result when I run the routine with the specified bounding box but without multilooking. Similarly, no errors result when I run the routine without specifying a bounding box but with multilooking. I have deleted and re-downloaded the products for that date, but the error persists. 

Has anyone encountered a similar issue in the past? Thanks for any assistance and happy to answer any questions.

ehavazli:
Hi @rzinke ,

Can you confirm the product is not corrupt, can you see the png of 20160903_20160717 in your unwrappedPhase folder?

rzinke:
Hi @ehavazli, the .png was generated and the image appears to be a valid, unwrapped interferogram.

ehavazli:
@rzinke, since you described it as a non-systematic error I am assuming it is an error related to the product. I would re-download that specific product and re-run ariaTSsetup.py again just to test. 

However, if it is not a key product for timeseries (e.g. keeping network connectivity), I would just discard it.

rzinke:
Agreed. 
I deleted and re-downloaded the products for that date, but the error still occurs.

I am continuing my data processing without that product, but I mostly wanted to put the word out there in case this issue is something more systematic. I'll send out the alert if I encounter it again.

ehavazli:
thanks for the feedback, I will look into it and try to reproduce the error.

sssangha:
@rzinke you had told me you also had this issue with a different pair (20160810_20160717). Just to clarify, are you having problem with both 20160810_20160717 and 20160903_20160717?

sssangha:
[...]

PR #130, opened 2020-01-17
Set resampling of unwphase to lanczos

sssangha:
Setting gdal resampling algorithm for downsampling of unwphase to lanczos (originally average) resolves a bug.

### DIFF
--- a/tools/ARIAtools/vrtmanager.py
+++ b/tools/ARIAtools/vrtmanager.py
@@ -151,7 +151,7 @@ def resampleRaster(fname, multilooking, bounds, prods_TOTbbox, rankedResampling=
         # Default: resample unw phase with gdal lanczos algorithm
         else:
             # Resample unwphase
-            gdal.Warp(fnameunw, fnameunw, options=gdal.WarpOptions(format=outputFormat, cutlineDSName=prods_TOTbbox, outputBounds=bounds, xRes=arrshape[0], yRes=arrshape[1], resampleAlg='average',multithread=True, options=['NUM_THREADS=%s'%(num_threads)+' -overwrite']))
+            gdal.Warp(fnameunw, fnameunw, options=gdal.WarpOptions(format=outputFormat, cutlineDSName=prods_TOTbbox, outputBounds=bounds, xRes=arrshape[0], yRes=arrshape[1], resampleAlg='lanczos',multithread=True, options=['NUM_THREADS=%s'%(num_threads)+' -overwrite']))
             # update VRT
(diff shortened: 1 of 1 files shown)


dbekaert:
@sssangha, @rzinke at some point some justification is needed on the impact of the kernel that gets applied and for us to understand the implications of the different types. Did we do a small sensitivity analysis? 

Perhaps @hfattahi and @piyushrpt might have some thoughts on this.

rzinke:
Hi @dbekaert, I am happy to help test the different kernels. Did you have any specific tests or deliverables in mind? I can switch the ResampAlg in gdal.Warp() from average to mode or lanczos. I can compare a few different interferograms, then run a full timeseries inversion with each to see if there are significant differences. Is that what you had in mind?

sssangha:
@dbekaert @rzinke ok, I have pushed an alternative work-around this gdal bug that allows us to maintain the average resampling for these fringe cases (so this commit applies to both the default multilooking and ranked resampling option).

A check is now implemented to confirm whether or not the unw phase file has been properly generated, and if it isn't then a negligible value (1e-6) is subtracted from the first pixel at the top left corner (i.e. array[0,0]). I still don't quite understanding why this is happening, but from my tests, it appears is a precision error only associated with this first entry in the array that leads to a failure in successfully writing the file. I have confirmed that this bug is not replicated on older gdal 2.x versions, so I will raise a gdal ticket. But for now, this is a stable solution.

As before, you can test this case out with the commands Rob had shared on the issues page (https://github.com/aria-tools/ARIA-tools/issues/129):
ariaDownload.py -t 41 -s 20160717 -e 20160903 -b '36.9 37.9 89.0 91.2'
#default resampling
ariaTSsetup.py -f 'products/*.nc' -w CroppedMultilooked -d download -b '36.9 37.9 89.0 91.2' -ml 3
#ranked resampling
ariaTSsetup.py -f 'products/*.nc' -w CroppedMultilooked_rr -d download -b '36.9 37.9 89.0 91.2' -ml 3 -rr

sssangha:
@rzinke can you please perform a sensitivity analysis on what I committed (i.e. average resampling) vs. lanczos as David suggested (so in the other test, just replace the average resampling method with lanczos)? E.g. distribution of stats (e.g. for each IFG average median, stdev, min/max). And how these properties do/don't vary with different multilooking factors?

rzinke:
@sssangha and @dbekaert, sure can. Will aim to have by end of day tomorrow.

dbekaert (approved):

PR #131, opened 2020-02-07
Solid Earth tide corrections implemented

sssangha:
Solid Earth tide corrections implemented from Dennis Milbert's source code (found on http://geodesyworld.github.io/SOFTS/solid.htm, and also leveraged by GMT's earthtide function).

To utilize, simply specify the "-st" flag when running either ariaExtract.py or ariaTSsetup.py, and the solid earth differential tides + corrected phase fields will be deposited under the "earthtidecorrected_products" subdirectory.

To install prerequiste fortran bindings, follow these steps:
1. Enter "solidearth_code" subdirectory under cloned ARIA-tools repo.
2. Generate python bindings: 'f2py -c -m solid_grid solid_grid.f'
3. Move generated "solid_grid.cpython*" bindings into your python site packages. To get site package location, enter 'python3 -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())"'

A slew of codacy suggested fixes also implemented, mainly breaking up long lines.

@rzinke, when you get back, it would be great for you to test this out over Tibet, where we have seen strong solid earth tide signal (relative to my tests over Afar). @bbuzz31, this may be of interest for you as well.

### DIFF
--- a/solidearth_code/solid_grid.f
+++ b/solidearth_code/solid_grid.f
@@ -1,0 +1,1613 @@
+      subroutine solidwrapped(iyr,imo,idy,ihh,imm,iss,glad1,glod1,
+     *                        glad2,glod2,steplon,steplat,increm,myfile)
+
+*** will compute 1 year of tide at given location starting at date hour 
+*** driver to test solid earth tide
+*** UTC version
+
+      implicit double precision(a-h,o-z), CHARACTER*512(m) 
+      dimension rsun(3),rmoon(3),etide(3),xsta(3)
+      logical lflag                    
+      !***^ leap second table limit flag
+      common/stuff/rad,pi,pi2
+      common/comgrs/a,e2
+
+*** constants
+
+      pi=4.d0*datan(1.d0)
+      pi2=pi+pi
+      rad=180.d0/pi
+
+*** grs80
+
+      a=6378137.d0
+      e2=6.69438002290341574957d-03
+
+      lout=1
+      open(lout,file=myfile,form='formatted',status='unknown')
+
+*** input section
+
+      if(iyr.lt.1901.or.iyr.gt.2099) then
+        write(*,'(*(i5,:,","))') 'error with input year=',iyr
+        go to 98
+      endif
+
+      if(imo.lt.1.or.imo.gt.12) then
+        write(*,'(*(i5,:,","))') 'error with input month=',imo
+        go to 98
+      endif
+
+      if(idy.lt.1.or.idy.gt.31) then
+        write(*,'(*(i5,:,","))') 'error with input day=',idy
+        go to 98
+      endif
+
+      if(ihh.gt.24) then
+        write(*,'(*(i5,:,","))') 'error with input hour=',ihh
+        go to 98
+      endif
+
+      if(imm.gt.60) then
+        write(*,'(*(i5,:,","))') 'error with input minute=',imm
+        go to 98
+      endif
+
+      if(iss.gt.60) then
+        write(*,'(*(i5,:,","))') 'error with input second=',iss
+        go to 98
+      endif
+
+      if(glad1.lt.-90.d0.or.glad1.gt.90.d0) then
+        write(*,'(*(G0.9,:,","))') 'error with input minlat=',glad1
+        go to 98
+      endif
+
+      if(glod1.lt.-360.d0.or.glod1.gt.360.d0) then
+        write(*,'(*(G0.9,:,","))') 'error with input minlon=',glod1
+        go to 98
+      endif
+
+      if(glad2.lt.-90.d0.or.glad2.gt.90.d0) then
+        write(*,'(*(G0.9,:,","))') 'error with input maxlat=',glad2
+        go to 98
+      endif
+
+      if(glod2.lt.-360.d0.or.glod2.gt.360.d0) then
+        write(*,'(*(G0.9,:,","))') 'error with input maxlon=',glod2
+        go to 98
+      endif
+
+*** header
+
+      write(lout,'(a,i5,2i3)') 'year/month/day= ',iyr,imo,idy
+      write(lout,'(a,i5,2i3)') 'hour/minute/second= ',ihh,imm,iss
+      write(lout,'(*(G0.9:," "))') 'steplon/steplat= ',steplon,steplat
+      write(lout,'(a)') 'lon,lat,ut,vt,wt'
+
+
+      nlat=((glad2-glad1)/steplat)                              
+      !***^ 10 grid points per degree
+      nlon=((glod2-glod1)/steplon)
+
+      do ilat=0,nlat,increm
+        do ilon=0,nlon,increm
+
+        glad=glad1+ilat*steplat
+        glod=glod1+ilon*steplon
+
+
+*** position of observing point (positive East)
+
+        if(glod.lt.  0.d0) glod=glod+360.d0
+        if(glod.ge.360.d0) glod=glod-360.d0
+
+        gla0=glad/rad
+        glo0=glod/rad
+        eht0=0.d0
+        call geoxyz(gla0,glo0,eht0,x0,y0,z0)
+        xsta(1)=x0
+        xsta(2)=y0
+        xsta(3)=z0
+
+
+*** here comes the sun  (and the moon)  (go, tide!)
+
+        ihr=   ihh
+        imn=   imm
+        sec=   iss                                       
+      !***^ UTC time system
+        call civmjd(iyr,imo,idy,ihr,imn,sec,mjd,fmjd)
+        call mjdciv(mjd,fmjd,iyr,imo,idy,ihr,imn,sec)    
+      !***^ normalize civil time
+        call setjd0(iyr,imo,idy)
+
+        lflag=.false.                                  
+      !***^ false means flag not raised
+        call sunxyz (mjd,fmjd,rsun,lflag)                   
+      !***^ mjd/fmjd in UTC
+        call moonxyz(mjd,fmjd,rmoon,lflag)                  
+      !***^ mjd/fmjd in UTC
+        call detide (xsta,mjd,fmjd,rsun,rmoon,etide,lflag)  
+      !***^ mjd/fmjd in UTC
+        xt = etide(1)
+        yt = etide(2)
+        zt = etide(3)
+
+*** determine local geodetic horizon components (topocentric)
+
+        call rge(gla0,glo0,ut,vt,wt,xt,   yt,   zt)       
+      !***^ tide vector
+
+        call mjdciv(mjd,fmjd               +0.001d0/86400.d0,
+     *              iyr,imo,idy,ihr,imn,sec-0.001d0)
+
+        write(lout,'(*(G0.9,:,","))') glod,glad,ut,vt,wt
+
+        enddo
+      enddo
+
+*** test flag and end of processing
+
+      if(lflag) then
+        write(*,'(a)') 'Mild Warning -- time crossed leap second table'
+        write(*,'(a)') '  boundaries.  Boundary edge value used instead'
+      endif
+
+      go to 99
+   98 write(*,'(a)') 'Check arguments ' 
+
+      return
+   99 end
+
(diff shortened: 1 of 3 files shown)


dbekaert:
This looks great and a welcomed contribution.

Wrt the installation i would recommend a few changes prior to merging such its in line with our current set-up.py capabilities.

>     1. Enter "solidearth_code" subdirectory under cloned ARIA-tools repo.
> 
>     2. Generate python bindings: 'f2py -c -m solid_grid solid_grid.f'
> 
>     3. Move generated "solid_grid.cpython*" bindings into your python site packages. To get site package location, enter 'python3 -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())"'

1) e.,g. We also have c-code programs for the 2stage unwrapper specific to ARIA-tools with permission of Author. See under bindings. If you have manually wrote code then that should be there too. 
2) Im not yet clear on the dependency from Dennis Milbert's source code. Are pieces of code copied or are we replicating his code base. I would recommend a similar structure as for the 2-stage unwrapped under thirdparty code. 
3) Setup.py should be updated if it finds the code that it compiles it. If there is a license agreement with the third party package, then we should not include it by default. (similar like for relaxIV)

bbuzzanga:
Cool @sssangha !
I will have a look at how it affects Hampton Roads.

sssangha:
> This looks great and a welcomed contribution.
> 
> Wrt the installation i would recommend a few changes prior to merging such its in line with our current set-up.py capabilities.
> 
> > ```
> > 1. Enter "solidearth_code" subdirectory under cloned ARIA-tools repo.
> > 
> > 2. Generate python bindings: 'f2py -c -m solid_grid solid_grid.f'
> > 
> > 3. Move generated "solid_grid.cpython*" bindings into your python site packages. To get site package location, enter 'python3 -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())"'
> > ```
> 
> 1. e.,g. We also have c-code programs for the 2stage unwrapper specific to ARIA-tools with permission of Author. See under bindings. If you have manually wrote code then that should be there too.
> 2. Im not yet clear on the dependency from Dennis Milbert's source code. Are pieces of code copied or are we replicating his code base. I would recommend a similar structure as for the 2-stage unwrapped under thirdparty code.
> 3. Setup.py should be updated if it finds the code that it compiles it. If there is a license agreement with the third party package, then we should not include it by default. (similar like for relaxIV)

sssangha:
Whoops, sorry about accidentally closing the PR, not sure how I managed to do that.

Anyways, @dbekaert the source code base must be replicated and converted to python bindings to work. My idea was for this code to be first reviewed for its performance, and once that is done and before we merge the code I will see how exactly to integrate this more seamlessly through setup.py. I am not so familiar with the protocol for integrating fortran source-codes through setup.py though, but we will figure it out.

And yes, since acknowledgements/permissions were not clearly articulated on his website, I sent an email out to the author of the source code. Will update the PR accordingly once I get his instructions/permission.

dbekaert:
one would would be to ask for the code to be put on github.

dbekaert:
Currently we have not much to benchmark it too.
Different flavors of algorithms could be possible.
I assume the python bindings could be your wrapper around the code.
Let us have a closer look into the replicating part next week.

bbuzzanga:
@sssangha can you clarify what the files refer to (_sediff vs no extension) and how you calculated the two files (not the actual earth tide correction)? Thanks.

sssangha:
@bbuzz31 "_sediff" refers to the solid earth tide differential, so similar naming convention to the tropospheric correction differential "_tropodiff" file generated in the "tropo_correction" function. The file generated without the extension is the corrected phase field (i.e. unw phase minus the differential), again consistent vis-a-vis the "tropo_correction" function.

To estimate "sediff", I just projected the ENU displacement (output of Milbert's fortran code) to LOS for each SAR date and then simply passed the difference. This difference is then interpolated to the full-resolution of the standard product, similar to what we do with the metadata layers under the "finalize_metadata" function. Makes sense?

bbuzzanga:
@sssangha great thanks, that's what I expected. I got confused in my comparisons by some strange (but very tiny) edge effects. You can seem them in the very bottom row. 
White values are 0, which they all should be.

edge_effects

dbekaert:
> @sssangha great thanks, that's what I expected. I got confused in my comparisons by some strange edge effects. You can seem them in the very bottom row.
> 
> image

for understanding: add labels for rows/columns and colorbar

bbuzzanga:
@dbekaert good call, that plot was wildly uninformative. Fixed in original post.

dbekaert:
> @dbekaert good call, that plot was wildly uninformative. Fixed in original post.
Thanks, that provides more context for @sssangha to fix. 

@sssangha with respect for a test, do we have a region where we know what the signal should look like and where we can use the model to estimate it. E.g. this to ensure we are not by accident adding versus removing the signal. 

@bbuzz31 the test over Norfolk will not have a significant impact as i suspect the solid earth tide to be of long-wavelength. Local referencing and small study area will remove the bulk of the solid earth tide (offset) and leave only a minor tilt.

dbekaert:
@rzinke perhaps it might be more applicable over a large region as yours. Although limited due to continental shield?

EJFielding:
I did a little look at the references for Dennis Milbert's source code, and I don't see any sign that they use a crustal or lithospheric thickness database in the solid Earth tide calculations, so it does not matter whether you do the calculation in a shield or ocean. The spatial length scale is the main factor in the size of the effect, along with the latitude.

rzinke:
Hi @dbekaert, I have tested the solid earth tide correction for an area of approximately one frame on the full series of data (all acquisitions of that area since 2014). The script appears to work quite well, though it is hard to gauge what effect the correction has. This is because the long-term trend of tidal measurements at a pixel only  result in a linear velocity of ~6 mm/yr based on the raw tidal values... I think this effect would be less if we use a reference point nearby, as we do for the timeseries calculation. I illustrate the solid earth tide values at each Track 41 epoch in the plot below:

Etide_fwdModel

I compared the timeseries velocity maps computed from the phase values for both non-S.E. tide corrected, and S.E. tide-corrected. 

FullResvsEtide_side-by-side

There is essentially no difference across the map, though that does not mean the correction itself isn't working, of course. 

Let me know how I else I can test this script. I would be happy to process a larger area, but then I will start running into memory issues as in the past.

dbekaert:
Correct the test would likely be more relevant for a large area.
Small area will be negligible based on a reference point

rzinke:
I am getting an error when using the solid earth tide correction. 

I run the command: 
ariaTSsetup.py -f "/Track019/AllProducts/*.nc" -w ML3_SE -d download -m download -b Shapefiles/19_B_Tibet_AOI.shp -nt 8 -ml 3 -mo 180000 -st

Which results in the error:
Rejected scene ML3_SE/productBoundingBox/20200315_20200303.shp has only 1338.083319km² overlap with bbox
Rejected scene ML3_SE/productBoundingBox/20200327_20200303.shp has only 1339.625020km² overlap with bbox
Rejected scene ML3_SE/productBoundingBox/20200327_20200315.shp has only 1339.625020km² overlap with bbox
Download/cropping DEM
Downloaded 3 arc-sec SRTM DEM here: ML3_SE/DEM/SRTM_3arcsec.dem

Extracting unwrapped phase, coherence, and connected components for each interferogram pair
Generating: unwrappedPhase - [==================================================] 20190607_20190526    0s /     0s
Generating: coherence - [==================================================] 20190607_20190526 3541s /    72s

Extracting single incidence angle, look angle and azimuth angle files valid over common interferometric grid
Generating: incidenceAngle - [==================================================] 20141218_20141031
Generating: lookAngle - [==================================================] 20141218_20141031
Generating: azimuthAngle - [==================================================] 20141218_20141031
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/bin/ariaTSsetup.py", line 15, in <module>
    main(inps)
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.6/site-packages/ARIAtools/tsSetup.py", line 370, in main
    outputFormat=inps.outputFormat, verbose=inps.verbose, num_threads=inps.num_threads)
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.6/site-packages/ARIAtools/extractProduct.py", line 944, in solidtide_correction
    yRes=abs(geotrans[-1]), resampleAlg='bilinear',multithread=True, options=['NUM_THREADS=%s'%(num_threads)+' -overwrite']))
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.6/site-packages/osgeo/gdal.py", line 613, in Warp
    srcDSTab = [Open(srcDSOrSrcDSTab)]
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.6/site-packages/osgeo/gdal.py", line 3308, in Open
    return _gdal.Open(*args)
RuntimeError: opj_read_header() failed


Perhaps this is a gdal issue?

sssangha:
@rzinke it's strange that it points to a gdal.Warp call on line 944 in "extractProduct.py" when that particular call is actually on line 913. So perhaps you made some modifications? I would stash any changes you made and make sure you are working off of this commit.

If the issue still persists, it could be because somehow the solid earth tide differential which is generated is empty. I would examine the output textfiles to confirm whether or not this is the case and perhaps then implement a tracer (i.e. import pdb ; pdb.set_trace()) to manually try to open the object with gdal and carry out other diagnostics.

tools/ARIAtools/extractProduct.py
```diff
@@ -341,8 +394,8 @@ def merged_productbbox(metadata_dict, product_dict, workdir='./', bbox_file=None
 
     return metadata_dict, product_dict, bbox_file, prods_TOTbbox, arrshape, proj
 
-
-def export_products(full_product_dict, bbox_file, prods_TOTbbox, layers, rankedResampling=False, dem=None, lat=None, lon=None, mask=None, outDir='./',outputFormat='VRT', stitchMethodType='overlap', verbose=None, num_threads='2', multilooking=None):
+def export_products(full_product_dict, bbox_file, prods_TOTbbox, layers, rankedResampling=False, dem=None, lat=None, lon=None, mask=None, \
```
dbekaert, line 397: Codacy found an issue: [export_products is too complex (17) (MC0001)](https://app.codacy.com/gh/aria-tools/ARIA-tools/pullRequest?prid=5789558)

dbekaert:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 1
- Added 1
           

Clones added
============
- tools/ARIAtools/tsSetup.py  5
- tools/ARIAtools/extractProduct.py  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/aria-tools/ARIA-tools/pullRequest?prid=5789558&bid=18951694)

solidearth_code/solid_grid.f
```diff
@@ -0,0 +1,1613 @@
+      subroutine solidwrapped(iyr,imo,idy,ihh,imm,iss,glad1,glod1,
+     *                        glad2,glod2,steplon,steplat,increm,myfile)
+
+*** will compute 1 year of tide at given location starting at date hour
```
yunjunz, line 4: Hi @sssangha, this is awesome! I am trying to understand the usage of this function: the output text file is the predicted solid earth tide for the given date/time, right? This comment is not in the original *.for file from Dennis. And I got confused a little bit.

sssangha: Hello @yunjunz, yes this file has been modified from the original source code. Originally it generated solid earth tide predictions for a given time and point location. I modified it to generate a coarse grid of predictions for a given time, which I then interpolate to the interferometric grid. 
Since I am concerned though that the source code will diverge, I aim to instead transition to wrapping around the GMT solid earth tide function (which uses the same source code). In this way, we may directly generate grids and be tied to GMT instead of diverging.

yunjunz: Thanks for the clarification @sssangha. Tieing with GMT does sound easier in the long term.

dbekaert:
@sssangha this can be retired now and point to the SET repo.

PR #140, opened 2020-04-15
ariaDownload option to return urls

bbuzzanga:
By using `--output Url`, this will now produce a file of urls (and also return them in a call to the class) that point to the requested GUNW products.
One such line of the file, or element in the returned list, could be https://grfn.asf.alaska.edu/door/download/S1-GUNW-A-R-004-tops-20190512_20190406-230629-37655N_35779N-PP-3ea0-v2_0_2.nc

The Url can then be accessed directly with GDAL in a python script via, e.g.

```python
from osgeo import gdal
gdal.SetConfigOption('GDAL_HTTP_COOKIEFILE','cookies.txt')
results = gdal.Info("/vsicurl/https://grfn.asf.alaska.edu/door/download/S1-GUNW-A-R-004-tops-20190512_20190406-230629-37655N_35779N-PP-3ea0-v2_0_2.nc")

### DIFF
--- a/tools/bin/ariaDownload.py
+++ b/tools/bin/ariaDownload.py
@@ -10,6 +10,7 @@
 
 import os
 import os.path as op
+import shutil
 import math
 import re
 import json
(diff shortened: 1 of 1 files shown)


dbekaert (approved):
@bbuzz31 thanks for the edits. I tested and confirm it works.
Thanks for the contribution.

PR #141, opened 2020-04-17
Circumvent netcdf4 driver to parse metadata

sssangha:
netCDF4 driver no longer used to open standard products in standard sorting class. However, devs may experiment with original, vestigial functions leveraging netCDF4 for now (i.e. "__OGmappingVersion__" and "__OGmappingData__")

Radar-metadata field "azimuthZeroDopplerMidTime" now parsed directly from filename, and also replaces references to "azimuthZeroDopplerStartTime" and "azimuthZeroDopplerEndTime" in the "extractProduct" and "tsSetup", since the latter two can only be accessed with the netCDF4 driver. 

This PR directly addresses suggestions made in issue #138

Issue #138, opened 2020-04-12
Support virtual file access


dbekaert:
Tracking in this issue ticket the discussion with changes needed to support virtual data access from ASF S3 bucket directly.

## ariaDownload.py - complete
~~Add option in ariaDownload.py to provide a list of where product patch wrt ASF S3.
Output should be able to be parsed into other aria*.py fucntions.~~ - **Complete**

## ARIAproduct.py - complete
~~Remove dependency on local netcdf reader. Following locations should be updated modified~~

1. product variable mapping: 
We could just hard-code the data-mapping for the different product versions and remove the need to leverage netcdf to search for the mapping. This should also speed up the process considerably at the cost of hard-coded code
 https://github.com/aria-tools/ARIA-tools/blob/98d72e77a84925d0205b551835c7d86efbd2dbac/tools/ARIAtools/ARIAProduct.py#L161

2. Pair-name retrieval:
Pair-name is retrieved from the netcdf using the original SLC granules for reference and secondary. 
Suggest replacing this with extracting from product name directly. 
https://github.com/aria-tools/ARIA-tools/blob/98d72e77a84925d0205b551835c7d86efbd2dbac/tools/ARIAtools/ARIAProduct.py#L154

3. Product without group support
Believe this can be removed: https://github.com/aria-tools/ARIA-tools/blob/98d72e77a84925d0205b551835c7d86efbd2dbac/tools/ARIAtools/ARIAProduct.py#L113

## Testing of the extract modules - in progress
The call to GDAL should directly work as long as the path handling is done correction. 
~~We could leverage S3 or vsi access: See examples here #27~~

Once above to are implemented would need some testing.

bbuzzanga:
#140  added an option in `ariaDownload.py` for returning urls to GUNW products at ASF that can be directly accessed via vsicurl

let me know if this isn't as expected

dbekaert:
@bbuzz31 ariaDownload.py with url option works as expected. 

Need to resolve #27 as conda build of GDAL recognized netcdf as h5 files for macos (and also linux in my case). However linux should be working. 
However a test assuming h5 seems to work to do a translate of amplitude out of the file:
```
from osgeo import gdal 
gdal.SetConfigOption('GDAL_HTTP_COOKIEFILE','cookies.txt') 
[...]

PR #143, opened 2020-04-22
Updates to handle virtual files

sssangha:
Added updates as proposed in issue #138 

Specifically, updates to appropriately load virtual products and also test whether or not the gdal install is properly configured to handle these virtual product files.

Additionally, files no longer opened with the gdal driver to parse product layer keys. These layer names are now hardcoded and parsed under the "__mappingVersion__" subfunction, similar to how radar metdata keys were handled in the previous commit #141

### DIFF
--- a/tools/ARIAtools/ARIAProduct.py
+++ b/tools/ARIAtools/ARIAProduct.py
@@ -51,10 +51,10 @@ class ARIA_standardproduct: #Input file(s) and bbox as either list or physical s
             self.files=[str(i) for i in filearg.split(',')]
             # If wildcard
             self.files=[os.path.abspath(item) for sublist in [self.glob.glob(os.path.expanduser(os.path.expandvars(i))) if '*' in i else [i] for i in self.files] for item in sublist]
-        elif os.path.basename(filearg).startswith('download'):
+        # If list of URLs provided
+        elif os.path.basename(filearg).endswith('.txt'):
             with open(filearg, 'r') as fh:
                 self.files = [f.rstrip('\n') for f in fh.readlines()]
-
         # If single file or wildcard
         else:
             # If single file
@@ -65,6 +65,17 @@ class ARIA_standardproduct: #Input file(s) and bbox as either list or physical s
                 self.files=self.glob.glob(os.path.expanduser(os.path.expandvars(filearg)))
             # Convert relative paths to absolute paths
             self.files=[os.path.abspath(i) for i in self.files]
+
+        # If URLs, append with '/vsicurl/'
+        self.files=['/vsicurl/{}'.format(i) if 'https://' in i else i for i in self.files]
+        # must configure gdal to load URLs
+        if any("https://" in i for i in self.files):
+            gdal.SetConfigOption('GDAL_HTTP_COOKIEFILE','cookies.txt')
+            #check if reader is being captured as netcdf
+            [s for s in self.files if 'https://' in s][0]
+            fmt=gdal.Open([s for s in self.files if 'https://' in s][0]).GetDriver().GetDescription()
+            if fmt != 'netCDF': raise Exception ('Your GDAL install is not configured to handle virtual files')
+
         if len(self.files)==0:
             raise Exception('No file match found')
         # If specified workdir doesn't exist, create it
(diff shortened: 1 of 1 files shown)


tools/ARIAtools/ARIAProduct.py
```diff
@@ -65,6 +65,17 @@ def __init__(self, filearg, bbox=None, workdir='./', verbose=False):
...
             # Convert relative paths to absolute paths
             self.files=[os.path.abspath(i) for i in self.files]
+
+        # If URLs, append with '/vsicurl/'
+        self.files=['/vsicurl/{}'.format(i) if 'https://' in i else i for i in self.files]
+        # must configure gdal to load URLs
+        if any("https://" in i for i in self.files):
+            gdal.SetConfigOption('GDAL_HTTP_COOKIEFILE','cookies.txt')
+            #check if reader is being captured as netcdf
+            [s for s in self.files if 'https://' in s][0]
```
dbekaert, line 75: Does this do anything here?

dbekaert: L75 might  have been used for debugging?

tools/ARIAtools/ARIAProduct.py
```diff
@@ -106,29 +117,23 @@ def __readproduct__(self, fname):
...
-            fname   = '/vsicurl/{}'.format(fname)
-            version ='URL'
-            fmt     = 'HDF5:"'
-        else:
-            try:
+        try:
+            #version accessed differently between URL vs downloaded product
+            if 'https://' in fname:
+                version=str(gdal.Open(fname).GetMetadata_Dict()['version'])
+            else:
```
dbekaert, line 127: Reading of 'NC_GLOBAL#version' should not be different if in S3.
L122-124 likely to be removed.
To be tested on opensarlabs as it needs the netcdf to be recognized!

tools/ARIAtools/ARIAProduct.py
```diff
@@ -65,6 +65,17 @@ def __init__(self, filearg, bbox=None, workdir='./', verbose=False):
...
+
+        # If URLs, append with '/vsicurl/'
+        self.files=['/vsicurl/{}'.format(i) if 'https://' in i else i for i in self.files]
+        # must configure gdal to load URLs
+        if any("https://" in i for i in self.files):
+            gdal.SetConfigOption('GDAL_HTTP_COOKIEFILE','cookies.txt')
+            #check if reader is being captured as netcdf
+            [s for s in self.files if 'https://' in s][0]
+            fmt=gdal.Open([s for s in self.files if 'https://' in s][0]).GetDriver().GetDescription()
+            if fmt != 'netCDF': raise Exception ('Your GDAL install is not configured to handle virtual files')
```
dbekaert: We should also add here a test for a single file which does not have https.
e.g. verify for a local file that it also gets captured with the netcdf reader.
If not netcdf error out and say version >2.3 is needed

dbekaert: L77 the error for S3 being h5, could be more specific indicating: Linux kernel >=4.3 and libnetcdf >=4.5

dbekaert:
@sssangha added a review with comments.

tools/ARIAtools/ARIAProduct.py
sssangha: Yup, will have to confirm on opensarlabs, once the technical issue on my end gets resolved. From my end locally though, I could successfully access the version from the URLs this way.

tools/ARIAtools/ARIAProduct.py
sssangha: Yup, will have to remove. Will do now.

tools/ARIAtools/ARIAProduct.py
sssangha: Yes, I will simplify and add in the proposed check for locally downloaded products.

sssangha:
@dbekaert I have made some of the suggested changes. Did not change the way the version is read yet though, as I cannot login to opensarlab yet to confirm.

dbekaert (approved):
looks good, Will merge it in and do a commit on top of this to make it work.
Will post on the issue ticket with more information,

PR #144, opened 2020-04-22
Rz aoi assist

rzinke:
Restructured ariaAOIassist program to exclude dates based on user input or semi-automated frame extent detection. This capability will allow the user to more efficiently create AOIs in Google Earth.

The programs now works with pandas dataframe class.

### DIFF
--- a/tools/bin/ariaAOIassist.py
+++ b/tools/bin/ariaAOIassist.py
@@ -1,38 +1,49 @@
-#!/usr/bin/env python3
+#!/anaconda3/envs/ARIA-tools/bin/python3
 
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# Author: Robert Zinke, Simran Sangha, David Bekaert
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+# Import modules
 import os
 import argparse
+import pandas as pd
 from datetime import datetime, timedelta
 import numpy as np
-import csv
-import itertools
-import pandas as pd
+import matplotlib.pyplot as plt
 from pandas.plotting import register_matplotlib_converters
 register_matplotlib_converters()
-from collections import OrderedDict
 from osgeo import ogr
 from shapely.geometry import Polygon
 
-import matplotlib as mpl
-# supress matplotlib postscript warnings
-mpl._log.setLevel('ERROR')
-import matplotlib.dates as mdates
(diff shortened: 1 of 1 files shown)


dbekaert:
@rzinke can you remove the thirdparty addition of files from the PR?

rzinke:
Hi @dbekaert, I reset my local branch and the new pull request includes changes only to the intended file. I'll keep an eye on the "Files changed" in the future.

EJFielding:
That was a lot of notifications this morning ;-).

tools/bin/ariaAOIassist.py
```diff
@@ -1,371 +1,505 @@
-#!/usr/bin/env python3
+#!/anaconda3/envs/ARIA-tools/bin/python3
```
dbekaert: this is local to your machine

rzinke: Right I noticed that, but it was written /anaconda3/... in the original code, so I must have used an older version of it. Good fix, thanks.

dbekaert: can you correct it to what it used to be?

rzinke: Yes, but I look at the way ariaAOIassist is currently written is the standard way, i.e., !/usr/bin/env python3. I should keep that, right?

dbekaert: yes that woudl be consistent with the other programs:
```
#!/usr/bin/env python3
```

dbekaert:
@rzinke would you be willing to make a small jupyter notebook to demo the functionality.
Talking a bit about the caveats on the bbox impacting the automatic generation of the AOI?

rzinke:
Absolutely. 
I'd also be interested if anyone has better solutions for trimming the AOI.
I'll send you what I have this afternoon.

dbekaert (approved):
@rzinke thanks for the contribution.

I have a few potential improvements that we can purse. 
e.g. I believe the shorting of the AOI with 20km, we could do an erosion operation of 20km.
This would also impact the width of the AOI, but on the other hand i do not believe that would be a cause for problems in the processing.

rzinke:
> @rzinke thanks for the contribution.
> 
> I have a few potential improvements that we can purse.
> e.g. I believe the shorting of the AOI with 20km, we could do an erosion operation of 20km.
> This would also impact the width of the AOI, but on the other hand i do not believe that would be a cause for problems in the processing.

This would likely be a more reliable approach than what I use currently. Do you have any ideas for how to erode the polygon?

rzinke:
> > @rzinke thanks for the contribution.
> > I have a few potential improvements that we can purse.
> > e.g. I believe the shorting of the AOI with 20km, we could do an erosion operation of 20km.
> > This would also impact the width of the AOI, but on the other hand i do not believe that would be a cause for problems in the processing.
> 
> This would likely be a more reliable approach than what I use currently. Do you have any ideas for how to erode the polygon?

This can be done with an affine transformation, no?

PR #145, opened 2020-04-25
Minor bug fixes

sssangha:
Fixed bug in ariaDownload that crashes when bbox specified. Filename of text file containing URLs updated so track name separated from index.

Removed hardcoded ionosphere layer key in ARIAProduct.py. The program would crash if the user specified an extraction of all layers (i.e. -l 'all') because the ionosphere layer doesn't exist. Checking for the existence of this layer would potentially re-introduce some overhead in the ARIAProduct script as 'gdalinfo' would have to be run on each product. For now, I proceed with the working assumption that future products containing this layer would have a distinct product version, with which we will associate the existence of this layer (i.e. update "__mappingVersion__" and "__mappingData__" in ARIAProduct.py accordingly)".

### DIFF
--- a/tools/ARIAtools/ARIAProduct.py
+++ b/tools/ARIAtools/ARIAProduct.py
@@ -263,7 +263,7 @@ class ARIA_standardproduct: #Input file(s) and bbox as either list or physical s
             # Layer names for these versions
             sdskeys=['productBoundingBox','/science/grids/data/unwrappedPhase','/science/grids/data/coherence',
             '/science/grids/data/connectedComponents','/science/grids/data/amplitude','/science/grids/imagingGeometry/perpendicularBaseline',
-            '/science/grids/imagingGeometry/parallelBaseline','/science/grids/imagingGeometry/incidenceAngle','/science/grids/imagingGeometry/lookAngle','/science/grids/imagingGeometry/azimuthAngle','/science/grids/imagingGeometry/ionosphere']
+            '/science/grids/imagingGeometry/parallelBaseline','/science/grids/imagingGeometry/incidenceAngle','/science/grids/imagingGeometry/lookAngle','/science/grids/imagingGeometry/azimuthAngle']
 
         return rdrmetadata_dict, sdskeys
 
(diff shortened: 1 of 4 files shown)


bbuzzanga:
Hi @sssangha could you give an example of the ariaDownload crashing for bbox? It seems to work for me. Also, the proposed changes result in files names like `download_products_-77W36S-75E38Nbbox_0.txt` which I find a bit cumbersome. Works fine though.

dbekaert:
I am confirming the same issue:

```
ValueError: time data '04:30:53' does not match format '%H:%M:%S.%f'
```

sssangha:
@bbuzz31 I know it isn't the prettiest naming convention, but at the same time I do want a way to record the bbox for a given download run. 

Speaking for myself, I often do play around and make tweaks to the bbox, so it would be useful to see the trade-offs (e.g. which products are introduced/left-out) by making adjustments to the bbox. Perhaps this may be just something for the user to sort out on their end, but it could be simpler just to include it in the filename as I proposed. What do you think?

Regarding the "bug", I double-checked and I believe I may have introduced my own bug that I resolved in this same PR. Specifically, I saw on line 182 that the filename was prepended with _ (e.g. the output filename would be '_/path/to/download_file.txt'), but in retrospect that doesn't seem to be in the original dev commit. However, I did have to make some changes to ariaDownload anyways because if the user specified both a track and bbox, the bbox would not have been appended to the filename. An if/elif/else statement had been imposed when writing the filename such that the track name would supersede references to the bbox. Does this make sense?

Also, good look on your exam!

bbuzzanga:
@sssangha yup sounds good. Sorry for the confusion on the bug, there was none and I had edited my comment to reflect my error. I'm guessing you just saw the email though.

And thanks!

sssangha:
@ehavazli this patch should now be consistent with prep_aria. I.e. there is no need to modify prep_aria.py to handle decimal vs absolute seconds.

dbekaert:
Tagging in @yunjunz on this too.

sssangha:
@ehavazli sorry, just fixed up some hanging chads in the other scripts

dbekaert:
@sssangha this might no longer be needed as merged through the circleCI PR

PR #153, opened 2020-04-27
KMZ now downloaded properly

sssangha:


sssangha:
@bbuzz31 as a follow-up to the email from @dbekaert, this wasn't a bug associated with the JSON url per se (which is needed to filter the products based off of user specified parameters), but rather a change I had introduced whereby the 'KML' keyword was no longer passed through to the call to 'requests'.

Seems to generate KMZs as expected now on my end, plus with dummy-checks for permutations of kml/kmz arguments (e.g. Kml, Kmz).

dbekaert (approved):
thnx for the fix

Issue #148, opened 2020-04-25
Ensuring raster sizes are the same


rzinke:
I have been reprocessing some data and I ran into an error related to the size of the rasters from ariaExtract. Specifically, when I run:
ariaExtract.py -w Extracted -f "products/*.nc" -d download -m download -croptounion -l all

I get the error:
Generating: bPerpendicular - [>                                                 ]Traceback (most recent call last):
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/bin/ariaExtract.py", line 16, in <module>
    main(inps)
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.6/site-packages/ARIAtools/extractProduct.py", line 757, in main
    export_products(standardproduct_info.products[1], standardproduct_info.bbox_file, prods_TOTbbox, inps.layers, inps.rankedResampling, dem=demfile, lat=Latitude, lon=Longitude, mask=inps.mask, outDir=inps.workdir, outputFormat=inps.outputFormat, stitchMethodType='overlap', verbose=inps.verbose, num_threads=inps.num_threads, multilooking=inps.multilooking)
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.6/site-packages/ARIAtools/extractProduct.py", line 396, in export_products
    finalize_metadata(outname, open_shapefile(bbox_file, 0, 0).bounds, prods_TOTbbox, dem, lat, lon, mask, outputFormat, verbose=verbose)
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.6/site-packages/ARIAtools/extractProduct.py", line 501, in finalize_metadata
    out_interpolated = mask.ReadAsArray()*out_interpolated
ValueError: operands could not be broadcast together with shapes (18481,7476) (18481,7477)

So it appears that when aria-tools resamples the data it does not ensure that all rasters (in this case the bPerpendicular) are the same size. Would you like me to try to build this fix in by specifying the array size?

sssangha:
@rzinke, hmm strange, I thought I had resolved this inconsistency a while back. Which revision is this though? Can you try again with my latest PR #145? 

If this issue persists with the latest PR, could you provide me with a list of products you had used so I can try to duplicate it?

EJFielding:
[...]

PR #156, opened 2020-04-29
Fixed interpolation bug in finalize_metadata

sssangha:
After investigating the problem outlined in Issue #148, I found that in some instances when an interpolated metadata layer is saved to the file, but then warped to apply a track cutline, rounding errors involved with the cutline may translate to 1 pixel in width/length with respect to the expected output dimensions. The program then crashes when a mask (with the expected output dimensions) is applied to this warped metadata layer.

Output width/lengths are now enforced through gdal.Warp in the finalize_metadata function, which ensures expected output dimensions are set.


rzinke (approved):
This fixed worked. Many thanks!

PR #159, opened 2020-05-03
Bb l cmask

bbuzzanga:
Create a mask using the [NLCD dataset](https://www.mrlc.gov/data?f%5B0%5D=category%3Aland%20cover&f%5B1%5D=year%3A2016). 

**You must have this downloaded.**

To create a mask using ARIA products you've already extracted (e.g. via `ariaExtract.py` or `ariaTSsetup.py`) run the `ariaNLCDmask.py` directly. See instructions with `ariaNLCDmask.py -h`

Alternatively, run ariaExtract.py` or ariaTSsetup.py` with `-m path/NLCD_2016_Land_Cover_L48_20190424.img` to create the mask on the fly. 

Finally, you can run `prep_aria.py --water-mask PATH` where PATH is the file produced by ariaNLCDmask.py 

Minimally tested, and takes a bit for large bounding boxes (~30 minutes for a whole track).

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -234,6 +234,10 @@ def prep_mask(product_dict, maskfilename, bbox_file, prods_TOTbbox, proj, amp_th
         del lake_masks, amp_file, mask_file
         os.remove(os.path.join(workdir,'watermsk_shorelines.vrt')); os.remove(os.path.join(workdir,'watermsk_lakes.vrt'))
 
+    if os.path.basename(maskfilename).lower().startswith('nlcd') and maskfilename.lower().endswith('img'):
+        from ariaNLCDmask import NLCDMasker
+        maskfilename = NLCDMasker(os.path.dirname(workdir))(maskfilename) ## write mask to disk
+
     # Load mask
     try:
         # Check if uncropped/cropped maskfiles exist in 'mask' subdirectory
@@ -249,8 +253,8 @@ def prep_mask(product_dict, maskfilename, bbox_file, prods_TOTbbox, proj, amp_th
             gdal.Translate(maskfilename+'.vrt', maskfilename, options=gdal.TranslateOptions(format="VRT"))
         #pass cropped DEM
         mask=gdal.Warp('', maskfilename, options=gdal.WarpOptions(format="MEM", cutlineDSName=prods_TOTbbox, outputBounds=bounds, width=arrshape[1], height=arrshape[0], multithread=True, options=['NUM_THREADS=%s'%(num_threads)]))
-        mask.SetProjection(proj)
-        mask.SetDescription(maskfilename)
+        # mask.SetProjection(proj)
+        # mask.SetDescription(maskfilename)
     except:
         raise Exception('Failed to open user mask')
 
@@ -340,7 +344,6 @@ def merged_productbbox(metadata_dict, product_dict, workdir='./', bbox_file=None
     del ds
 
     return metadata_dict, product_dict, bbox_file, prods_TOTbbox, arrshape, proj
-
 
 def export_products(full_product_dict, bbox_file, prods_TOTbbox, layers, rankedResampling=False, dem=None, lat=None, lon=None, mask=None, outDir='./',outputFormat='VRT', stitchMethodType='overlap', verbose=None, num_threads='2', multilooking=None):
     """
@@ -695,7 +698,6 @@ def tropo_correction(full_product_dict, tropo_products, bbox_file, prods_TOTbbox
         else:
             print("WARNING: Must skip IFG %s, because the tropospheric products corresponding to the reference and/or secondary products are not found in the specified folder %s"%(product_dict[2][i][0],tropo_products))
 
-
 def main(inps=None):
     '''
         Main workflow for extracting layers from ARIA products
@@ -741,10 +743,6 @@ def main(inps=None):
     # extract/merge productBoundingBox layers for each pair and update dict,
     # report common track bbox (default is to take common intersection, but user may specify union), and expected shape for DEM.
     standardproduct_info.products[0], standardproduct_info.products[1], standardproduct_info.bbox_file, prods_TOTbbox, arrshape, proj = merged_productbbox(standardproduct_info.products[0], standardproduct_info.products[1], os.path.join(inps.workdir,'productBoundingBox'), standardproduct_info.bbox_file, inps.croptounion, num_threads=inps.num_threads, minimumOverlap=inps.minimumOverlap)
-    # Load or download mask (if specified).
-    if inps.mask is not None:
-        inps.mask = prep_mask([[item for sublist in [list(set(d['amplitude'])) for d in standardproduct_info.products[1] if 'amplitude' in d] for item in sublist], [item for sublist in [list(set(d['pair_name'])) for d in standardproduct_info.products[1] if 'pair_name' in d] for item in sublist]], inps.mask, standardproduct_info.bbox_file, prods_TOTbbox, proj, amp_thresh=inps.amp_thresh, arrshape=arrshape, workdir=inps.workdir, outputFormat=inps.outputFormat, num_threads=inps.num_threads)
-
 
     # Download/Load DEM & Lat/Lon arrays, providing bbox, expected DEM shape, and output dir as input.
     if inps.demfile is not None:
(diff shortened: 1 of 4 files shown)


EJFielding:
Sounds like this could be a big help in areas with wetlands that have some coherence but wild phase variations due to water level changes. I am not working on that kind of area at the moment, but I have seen it before. The high coherence but large phase variations is even worse at L-band.

tools/ARIAtools/extractProduct.py
```diff
@@ -249,8 +253,8 @@ def prep_mask(product_dict, maskfilename, bbox_file, prods_TOTbbox, proj, amp_th
             gdal.Translate(maskfilename+'.vrt', maskfilename, options=gdal.TranslateOptions(format="VRT"))
         #pass cropped DEM
         mask=gdal.Warp('', maskfilename, options=gdal.WarpOptions(format="MEM", cutlineDSName=prods_TOTbbox, outputBounds=bounds, width=arrshape[1], height=arrshape[0], multithread=True, options=['NUM_THREADS=%s'%(num_threads)]))
-        mask.SetProjection(proj)
-        mask.SetDescription(maskfilename)
+        # mask.SetProjection(proj)
+        # mask.SetDescription(maskfilename)
```
dbekaert, line 257: Any reason why you hash this out?

tools/ARIAtools/extractProduct.py
```diff
@@ -753,6 +751,10 @@ def main(inps=None):
     else:
         demfile=None ; Latitude=None ; Longitude=None
 
+    # Load or download mask (if specified).
```
dbekaert, line 754: Any specific reason why order is swapped with dem check?

dbekaert (changes requested):
@bbuzz31 thanks for the new functionality.

Not sure this would work start to end.
1. You do an import from the bin folder which is not set as part of the pythonpath during setup.py
    Can you add in a equivalent fucntion in the python folder, and like the other programs except (ariaDownload) call the python and only leave an argparse in the bin folder.
2. can you add in a small unit test to see it works. (add this to the .circleCI/*.cfg folder)
3. few changes which i do not think are changes, just a change in order.

tools/ARIAtools/tsSetup.py
```diff
@@ -251,16 +251,17 @@ def main(inps=None):
...
-    if inps.mask is not None:
-        inps.mask = prep_mask([[item for sublist in [list(set(d['amplitude'])) for d in standardproduct_info.products[1] if 'amplitude' in d] for item in sublist], [item for sublist in [list(set(d['pair_name'])) for d in standardproduct_info.products[1] if 'pair_name' in d] for item in sublist]],inps.mask, standardproduct_info.bbox_file, prods_TOTbbox, proj, amp_thresh=inps.amp_thresh, arrshape=arrshape, workdir=inps.workdir, outputFormat=inps.outputFormat, num_threads=inps.num_threads)
 
     # Download/Load DEM & Lat/Lon arrays, providing bbox, expected DEM shape, and output dir as input.
     if inps.demfile is not None:
         print('Download/cropping DEM')
         # Pass DEM-filename, loaded DEM array, and lat/lon arrays
         inps.demfile, demfile, Latitude, Longitude = prep_dem(inps.demfile, standardproduct_info.bbox_file, prods_TOTbbox, proj, arrshape=arrshape, workdir=inps.workdir, outputFormat=inps.outputFormat, num_threads=inps.num_threads)
 
+    # Load or download mask (if specified).
```
dbekaert, line 261: Looks like an order change. please confirm

tools/bin/ariaNLCDmask.py
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env python3
```
dbekaert, line 1: move this code as pythin code and keep a bin equivalent that only has an argparse. e.g. ariaTSprep.py, ariaExtract.py

tools/ARIAtools/extractProduct.py
```diff
@@ -234,6 +234,10 @@ def prep_mask(product_dict, maskfilename, bbox_file, prods_TOTbbox, proj, amp_th
         del lake_masks, amp_file, mask_file
         os.remove(os.path.join(workdir,'watermsk_shorelines.vrt')); os.remove(os.path.join(workdir,'watermsk_lakes.vrt'))
 
+    if os.path.basename(maskfilename).lower().startswith('nlcd') and maskfilename.lower().endswith('img'):
+        from ariaNLCDmask import NLCDMasker
```
dbekaert, line 238: This is a python import while the code is binary.

dbekaert (commented):
Where needed can you update the argparse with this input option for the enhanced masking, e.g. ariaExtract, ariaTSprep

tools/ARIAtools/extractProduct.py
bbuzzanga: nope not sure how that happened

tools/ARIAtools/extractProduct.py
bbuzzanga: this is important because my code uses the DEM to resample the NLCD product.

tools/ARIAtools/tsSetup.py
bbuzzanga: yes, see previous

tools/ARIAtools/extractProduct.py
dbekaert: This might be problematic for ariaExtract.py as dem is not a required input. What do you need? Proj and transformation?

tools/ARIAtools/extractProduct.py
```diff
@@ -753,6 +751,10 @@ def main(inps=None):
     else:
         demfile=None ; Latitude=None ; Longitude=None
 
+    # Load or download mask (if specified).
+    if inps.mask is not None:
+        inps.mask = prep_mask([[item for sublist in [list(set(d['amplitude'])) for d in standardproduct_info.products[1] if 'amplitude' in d] for item in sublist], [item for sublist in [list(set(d['pair_name'])) for d in standardproduct_info.products[1] if 'pair_name' in d] for item in sublist]], inps.mask, standardproduct_info.bbox_file, prods_TOTbbox, proj, amp_thresh=inps.amp_thresh, arrshape=arrshape, workdir=inps.workdir, outputFormat=inps.outputFormat, num_threads=inps.num_threads)
+
     # Extract user expected layers
     export_products(standardproduct_info.products[1], standardproduct_info.bbox_file, prods_TOTbbox, inps.layers, inps.rankedResampling, dem=demfile, lat=Latitude, lon=Longitude, mask=inps.mask, outDir=inps.workdir, outputFormat=inps.outputFormat, stitchMethodType='overlap', verbose=inps.verbose, num_threads=inps.num_threads, multilooking=inps.multilooking)
```
sssangha, line 759: While it looks like this wasn't added in during this particular PR, clearly the Latitude/Longitude inputs are not used. Perhaps they are just vestiges so they should be removed altogether, along with the prep_dem call where they are created

sssangha: Actually, now I see how they are used in the "finalize_metadata" function. I think it is fine to pass as so and cleaner to generated in the "prep_dem" function as is, however we should probably give different variable names in case there are conflicts with other python modules (e.g. how about lats/lons instead?)

sssangha:
@bbuzz31 do you think maybe in light of these proposed changes it would be more appropriate to just move the entire "prep_mask" function to a separate script and have a separate script dedicated to just generating the masks? You should still of course keep the "NLCDMasker" class separate as it is meant to leverage the NLCD dataset. You think this may be straightforward to implement for this PR, or more complicated?

tools/ARIAtools/extractProduct.py
sssangha: If you need just the projection, why not pass it from the "merged_productbbox" function (see line 343 of extractProduct.py)? It is one of those functions outputs

sssangha: Plus to warp it to have the same expected dimensions/resolution as the dem, you could pass a lot of the same variables as we do from "merged_productbbox" to "prep_dem" (e.g. see the way "arrshape" from "merged_prodduct" is used in the gdal.Warp commands in "prep_dem")

bbuzzanga:
This is continued at https://github.com/aria-tools/ARIA-tools/pull/169

PR #160, opened 2020-05-06
tsSetup: use relative path to each data vrt file in stack vrt file

yunjunz:
This PR replaces the absolute file path of the individual data vrt file in the stack/*Stack.vrt files with the relative path, so that `prep_aria.py` would still work if one moved the root directory.


dbekaert:
@yunjunz it fails a unit test where it calls prep_aria.py.
Are you suggesting to change the unit test to allow this behavior?
See here: https://github.com/aria-tools/ARIA-tools/blob/a9de4fc5e7bda8dff04f6056ff7c43a7f8e6bfa5/.circleci/config.yml#L81

yunjunz:
Hi @dbekaert, it turns out we need to add `relativeToVRT="1"` to allow for this behavior, as described here: https://gdal.org/drivers/raster/vrt.html#simplesource.

sssangha:
@dbekaert looks good to me, thanks @yunjunz!

PR #162, opened 2020-05-06
fix imshow interp issue

bbuzzanga:
imshow antialiasing;
see: https://github.com/matplotlib/matplotlib/issues/17345

### DIFF
--- a/tools/ARIAtools/ARIAProduct.py
+++ b/tools/ARIAtools/ARIAProduct.py
@@ -310,14 +310,20 @@ class ARIA_standardproduct: #Input file(s) and bbox as either list or physical s
         track_rejected_pairs=[]
 
         # Check for (and remove) duplicate products
+        num_prods=len(self.products)
+        num_dups=[]
         for i in enumerate(self.products[:-1]):
             # If scenes share >90% spatial overlap AND same dates, they MUST be duplicates. Reject the latter.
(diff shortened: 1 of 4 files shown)


dbekaert (commented):
Do you have an example on the impact of the plotting?
Do i understand correctly that with "none" it still is randomly picking pixels?

bbuzzanga:
Here are examples output from `ariaPlot.py`. The top plot the flag `interpolation='none'` set.
The bottom has no flag set.

Screen Shot 2020-05-06 at 19 14 11
Screen Shot 2020-05-06 at 19 13 50

Could you clarify about randomly picking pixels - is that in reference to the matplotlib issue?

dbekaert:
@bbuzz31 correct in the issue ticket you made on matlplotlib site. With the interpolation method to none we are now using all pixels or is this a prefixed image rendering resolution?

yunjunz:
I would recommend checking `interplation=nearest` as well, because this antialiasing effect is usually more severe in the active display window than in the saved image files, as we could save the figure with higher dpi than the resolution of our monitor. For example, view.py is using 300 by default. I am using  >= 600 dpi for all my publication figures. With 600 dpi and a figure size of 5 by 3 inches, this gives 3000 by 1800 pixels, which might be larger than the actual data size.

In these circumstances, as described in the [matplotlib doc](https://matplotlib.org/gallery/images_contours_and_fields/image_antialiasing.html#sphx-glr-gallery-images-contours-and-fields-image-antialiasing-py), upsampling is desired and `interpolation = nearest` is more appropriate as we could see the sharp boundary between pixels.

bbuzzanga:
thanks for the tip @yunjunz I've changed the interpolation in `ariaPlot.py` to `nearest` accordingly. @dbekaert I think @yunjunz comments answered your question?

dbekaert (approved):
@bbuzz31 looks good.

PR #164, opened 2020-05-09
New checks for minimum threshold option + verbose mode

sssangha:
Program exists if specified minimum threshold area supersedes area of user-specified bounding box. 

Various warning messages and notices of rejected scenes now segregated to verbose mode.

Addresses points raised in issue #161


dbekaert (approved):
Looks good. Thanks for the fixes!

PR #166, opened 2020-05-11
Fixed bug in finalize_metadata function. VRT now updated

sssangha:
Extracted/interpolated metadata field VRT file now updated to reflect cropping of metadata field raster. Not updating the VRT files could potentially lead to failures later on when trying to load the raster from the VRT file (e.g. because of inconsistent dimensions, nodata values, etc). 

For example, see note the inconsistency in reported dimensions before the bug-fix: 
Screen Shot 2020-05-11 at 1 51 32 AM
Screen Shot 2020-05-11 at 1 51 43 AM

Which is eliminated after the bug-fix:
Screen Shot 2020-05-11 at 1 52 57 AM
Screen Shot 2020-05-11 at 1 53 05 AM


dbekaert (approved):
verified by Cherl

PR #167, opened 2020-05-20
Rz misclosure

rzinke:
Adapted computeMisclosure script work with unwrapStack.vrt file. Numerous improvements to the functionality and readability of the code implemented. Documentation and examples updated.


dbekaert:
@rzinke the two-stage third-party dependency is creeping back in.

PR #169, opened 2020-05-27
Bb l cmask3

bbuzzanga:
_Continuation of https://github.com/aria-tools/ARIA-tools/pull/159_

Create a mask using the [NLCD dataset](https://www.mrlc.gov/data?f%5B0%5D=category%3Aland%20cover&f%5B1%5D=year%3A2016). 

~~You must have this downloaded.~~ Can automatically download with `-m nlcd`

To create a mask using ARIA products you've already extracted (e.g. via `ariaExtract.py` or `ariaTSsetup.py`) run the `ariaNLCDmask.py` directly. See instructions with `ariaNLCDmask.py -h`

Alternatively, run ariaExtract.py` or ariaTSsetup.py` with `-m path/NLCD_2016_Land_Cover_L48_20190424.img` to create the mask on the fly. 

Finally, you can run `prep_aria.py --water-mask PATH` where PATH is the file produced by ariaNLCDmask.py 

Minimally tested, and takes a bit for large bounding boxes (~30 minutes for a whole track).

All masking functions moved to separate script.


bbuzzanga:
I've implemented function to automatically download the NLCD mask and crop it over the requested region by passing `mask=nlcd'. There is also a unit test now.

dbekaert:
CircleCI seems to report a failure for the unit-test

ehavazli:
```
***Extract Product Function:***
Multi-core version
All (2) GUNW products meet spatial bbox criteria.
Group GUNW products into spatiotemporally continuous interferograms.
All (1) interferograms are spatially continuous.
No layers specified; only creating bounding box shapes
Thread count specified for gdal multiprocessing = 2
***Downloading and cropping the NLCD mask...***
Warning 1: for band 1, destination nodata value has been clamped to 255, the original value being out of range.

Received "killed" signal
```
the data is too large to be downloaded and handled/cropped?

bbuzzanga:
A brief web search suggests the killed signal error is related to memory which agree with your thoughts @ehavazli. 
This is my first use with .circle ci though I'm not sure how at the moment how to test it

dbekaert:
@bbuzz31 how much of data are you downloading? Circleci has indeed limited memory and can time-out.

bbuzzanga:
@dbekaert on the server the NLCD dataset is about ~20 GB. It's read into memory by gdal so must be chunked somehow; I have 8 GB of RAM on my computer and it works fine. All that's written to disk is the cropped mask. For one sentinel frame it's about 15 MB.

dbekaert:
CirclecI has only limited ram and memory under the free mode. I suspect it would really need to be a viscurl or vrt access. @bbuzz31 you mentioned before the problem was with credentials or permission to access it virtually?

bbuzzanga:
@dbekaert that problem fixed itself somehow; I am accessing the landcover mask directly with vsi curl (https://github.com/bbuzz31/ARIA-tools/commit/b719096db8c87dd26a8c9f5686912a024296c060#L135; https://github.com/bbuzz31/ARIA-tools/commit/b719096db8c87dd26a8c9f5686912a024296c060#L197)

dbekaert:
@bbuzz31 what about doing it like this?         gdal.BuildVRT(os.path.join(workdir,'SRTM_3arcsec_uncropped.dem.vrt'), _world_dem, options=gdal.BuildVRTOptions(outputBounds=bounds))

bbuzzanga:
@dbekaert actually building the VRT isn't necessary; you can directly warp / translate on the /vsi/url. Currently I'm fixing the first dumb problem: the downloaded products are over hawii - the mask is only lower 48

bbuzzanga:
Can someone provide information on how to get circle ci to read the .netrc that's created? it works fine for track 124 but not track 004 ??

dbekaert:
@bbuzz31 we leveraged download directly from the ARIA S3 bucket. We do not download directly from ASF. Can you search for the exact product using aria-products page and then use the single product download underneath the product info to get the ARIA S3 bucket url?

 I have figured out how to pass credentials in circleCI as variables. One should not include them in the circle code, but should be set as configurable parameters in its online settings which are encrypted. When i find some time i will add an account into that and we can replace with DAAC download options.

dbekaert:
> @dbekaert actually building the VRT isn't necessary; you can directly warp / translate on the /vsi/url. Currently I'm fixing the first dumb problem: the downloaded products are over hawii - the mask is only lower 48

The exit error was related to it not being covered or memory?
 @sssangha thoughts as well?

sssangha:
@dbekaert from my own tests on opensarlabs, doesn't seem to be any memory error. I went ahead and pushed an update where I download the products directly from the bucket. If it works, I will clean up circleci to reincorporate the existing checks.

sssangha:
@bbuzz31 Ok so I confirmed that the code exits out because it runs out of memory "code exit status 137". It's strange because usually the opensarlabs seems to be a good baseline in that it is also more conservative in terms of the amount of RAM allocated, and it seems to run through fine.

bbuzz31:
the code is breaking here `        dst = gdal.GetDriverByName('ENVI').Create(path, width, height, 1, gdalconst.GDT_Float32)`

tools/ARIAtools/mask_util.py
```diff
@@ -0,0 +1,306 @@
...
+            ds_out.GetRasterBand(i+1).SetNoDataValue(noData)
+    else:
+        # ds_out = gdal.GetDriverByName('MEM').Create('', ds_orig.RasterXSize,
+        #             ds_orig.RasterYSize, 1, gdal.GDT_Float32)
+        dst = os.path.join(os.getcwd(), 'tmp')
+        print ('here')
+        ds_out = gdal.GetDriverByName('ENVI').Create(dst, ds_orig.RasterXSize,
+                    ds_orig.RasterYSize, 1, gdal.GDT_Float32)
+        print ('created dataset')
+        ds_out.GetRasterBand(1).WriteArray(arr)
```
dbekaert, line 186: @bbuzz31 if i understand you mentioned offline that the problem of the exit code is happening here?
Any thoughts @sssangha?

bbuzzanga: yup; line 196. I think I will have to edit most of the workflow to only use VRTs

sssangha: @bbuzz31 one easy way to debug the code would be to use 'tracemalloc' (see option 2 here: https://medium.com/survata-engineering-blog/monitoring-memory-usage-of-a-running-python-program-49f027e3d1ba). Just start it before line 196, and perhaps implement a try/except block such that when it runs out of memory, you will print out the traced memory.
E.g. something like this:
```
import tracemalloc
tracemalloc.start()
try:
    ds_out.GetRasterBand(1).WriteArray(arr)
except:
    current, peak = tracemalloc.get_traced_memory()
    print(f"Current memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB")
    tracemalloc.stop()
    raise Exception('ran out of memory')
```

bbuzzanga:
Thanks will do. I had done a quick search for something like this and hadn’t found anything nearly as useful as this 👍🏽

Issue #137, opened 2020-04-07
Discontinuity in the azimuth angle (track-087)


yuankailiu:
**Type:** Might be a Bug report
**About:** Extraction of imaging geometry layers: Discontinuity in the azimuth angle
**Tools**: ARIA-tools, MintPy

---

**Description of the problem**

After downloading the standard products from track-087 (ascending) in ARIA, I used MintPy to extract layers and prepared time-series analysis. [To load the ARIA interferograms into MintPy](https://nbviewer.jupyter.org/github/insarlab/MintPy-tutorial/blob/master/smallbaselineApp_aria.ipynb), I ran the `prep.py` command in MintPy. 

By running this command, the "inputs" directory inside the working directory is created, and two HDF5 files are produced. One of them is `geometryGeo.h5`, which contains geometrical datasets including incidence angle, azimuth angle, shadow layover mask, slant range distance, and height.

Then I plotted the geometry using MintPy command, `view.py`. Here is the figure, which shows a sharp discontinuity in the azimuth angle (parallel to the latitude). I think this is not expected behavior and maybe shows an issue somewhere?

![image](https://user-images.githubusercontent.com/55262505/78727203-de90d180-78e8-11ea-9587-64f65d36262d.png)

Below is a close look at the azimuth angle here (a weird jump at latitude~N27.2 degree):

![image](https://user-images.githubusercontent.com/55262505/78728944-97590f80-78ed-11ea-898e-4c3a166ac397.png)

I did the same procedure for track-160, but the azimuth angle looks fine.

![image](https://user-images.githubusercontent.com/55262505/78728527-6debb400-78ec-11ea-8cf1-69711d9121d3.png)

**Full code that generated the `geometryGeo.h5` file**

```
prep_aria.py -s ../ariaGUNW_087/stack/ -i ../ariaGUNW_087/incidenceAngle/20141023_20141011.vrt -d ../ariaGUNW_087/DEM/SRTM_3arcsec.dem -a ../ariaGUNW_087/azimuthAngle/20141023_20141011.vrt --water-mask ../ariaGUNW_087/mask/watermask.msk
```
.

**Full message after running this code**

```
extract metadata from /net/kraken/nobak/ykliu/Aqaba_processing/ariaGUNW_087/stack/unwrapStack.vrt
--------------------------------------------------
create HDF5 file /net/kraken/nobak/ykliu/Aqaba_processing/mintpy_A087/inputs/geometryGeo.h5 with w mode
[...]

PR #170, opened 2020-05-27
Circumvent bPerp grid artifacts

sssangha:
In some cases, there is an abrupt sign-change in the perpendicular baseline grids (see ISCE issue ticket #137).    
Screen Shot 2020-05-27 at 1 31 35 AM

As a post-processing fix, I have implemented a strategy to first capture this artifact by plotting a best fit line to values along the mid-range and mid-azimuth, respectively. These plots are saved in verbose mode and organized as so in the specified working directory, respectively: 'metadatalyr_plots/bPerpendicular/20181210_20171203_midrange.eps' and 'metadatalyr_plots/bPerpendicular/20181210_20171203_midazimuth.eps'.    
Screen Shot 2020-05-27 at 1 34 14 AM

The covariance of the polynomial coefficient estimates between both plots are then passed and if either is larger than a hardcoded value of 0.001, then the correction is applied. Note that generally smoothly varying fields with no artifacts are associated with covariances on the order of 10-5, so this hardcoded value of 0.001 is appropriate. In regards to the correction procedure, a uniform sign is set to the entire field based off of whatever the majority sign is (e.g. if >50% of the pixels are negative, than the whole field is set to negative, and vice-versa), so it is assumed that whatever is in the majority is accurate. The logic behind this strategy is to not capture false positives for which a transition from positive to negative values is expected in a smoothly varying field (e.g. range of 0.1 m to -0.1 m), as opposed to fields with legitimate artifacts (e.g. range of 8 m to -8 m). Below is the corrected field:
Screen Shot 2020-05-27 at 1 33 25 AM

To replicate this test, run the following commands:
ariaDownload.py --track 121 -o download -i '20181210_20171203' -b '38.6 38.7 88.8 88.9'

ariaExtract.py -f "products/S1-GUNW-D-R-121-tops-20181210_20171203-000743-39338N_38026N-PP-eb7f-v2_0_2.nc" -l 'bPerpendicular' -d download -w testbugfix -verbose

### DIFF
--- a/tools/ARIAtools/ARIAProduct.py
+++ b/tools/ARIAtools/ARIAProduct.py
@@ -136,16 +136,15 @@ class ARIA_standardproduct: #Input file(s) and bbox as either list or physical s
 
         # Open standard product bbox
         if self.bbox is not None:
-            file_bbox = open_shapefile(fname + '":'+sdskeys[0], 'productBoundingBox', 1)                    ##SS => We should track the projection of the shapefile. i.e. in case this changes in the product
-            # file_bbox = open_shapefile(fname, 'productBoundingBox', 1)                    ##SS => We should track the projection of the shapefile. i.e. in case this changes in the product
+            file_bbox = open_shapefile(fname + '":'+sdskeys[0], 'productBoundingBox', 1)
             # Only generate dictionaries if there is spatial overlap with user bbox
             if file_bbox.intersects(self.bbox):
-                product_dicts = [self.__mappingData__(fname, rmdkeys, sdskeys)]
+                product_dicts = [self.__mappingData__(fname, rmdkeys, sdskeys, version)]
             else:
                 product_dicts = []
         # If no bbox specified, just pass dictionaries
         else:
-            product_dicts = [self.__mappingData__(fname, rmdkeys, sdskeys)]
+            product_dicts = [self.__mappingData__(fname, rmdkeys, sdskeys, version)]
 
         return product_dicts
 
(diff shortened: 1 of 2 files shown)


sssangha:
@rzinke I may have mentioned this in passing, but this particular issue is of direct relevance for one of your Tibet tracks!

sssangha:
Metadata quality control function now modified and moved into a separate class called ‘metadata_qualitycheck’.

The following 3 documented artifacts are flagged and corrected: (1) Large offset sign-flip in perpendicular baseline file, (2) Subtle artifacts, and (3) Large offsets across frame boundaries. 

Examples corresponding to each case are outlined below .

(1) Large offset sign-flip in perpendicular baseline file. This is resolved by determining which sign the majority of pixels possess, and forcing all pixels to share the same sign. Note that cases for which a sign-flip is expected (e.g. in smoothly varying fields that range from 0.1 to -0.1) are not flagged, as R^2 and the standard error of the best-fit line should be sufficiently high and low, respectively.
Replicate as so:
```ariaDownload.py --track 121 -o download -i '20181210_20171203' -b '38.6 38.7 88.8 88.9'```
```ariaExtract.py -nt 30 -f "/u/sar-r2/ssangha/ARIA-product_bug/bPerp_bug/T121/products/S1-GUNW-D-R-121-tops-20181210_20171203-000743-39338N_38026N-PP-eb7f-v2_0_2.nc" -l 'bPerpendicular' -d download -w ARIAprodrun_testbugfix_T121 -of ISCE -verbose```
Screen Shot 2020-05-30 at 4 01 19 PM


(2) Subtle artifacts. This is resolved by passing best fit linear plane to each affected raster band.
Replicate as so:
```ariaDownload.py --track 144 -o download -i '20150612_20150519' -b '39.8 40 -118.6 -118 '```
```ariaExtract.py -nt 30 -f "/u/sar-r2/ssangha/ARIA-product_bug/inc_bug/T144/products/S1-GUNW-D-R-144-tops-20150612_20150519-135847-40882N_38909N-PP-e3e6-v2_0_2.nc" -l 'bPerpendicular, bParallel, incidenceAngle, lookAngle, azimuthAngle' -d download -w ARIAprodrun_testbugfix_T144 -of ISCE -verbose```
Screen Shot 2020-05-30 at 4 14 53 PM


(3) Large offsets across frame boundaries. A best-fit plane to the original affected raster band would likely be very unreliable, so outlier values must be masked. To accomplish this, all values above or below the mean of the raster band are masked, depending on whichever is in the minority.
Replicate as so:
```ariaDownload.py --track 87 -o download -i '20151111_20150912' -b '28 29.5 34.2 37.4'```
```ariaExtract.py -nt 30 -f "/u/sar-r2/ssangha/ARIA-product_bug/az_angle_bug/T087/products/S1-GUNW-A-R-087-tops-20151111_20150912-153927-29028N_26970N-PP-0a7b-v2_0_2.nc,/u/sar-r2/ssangha/ARIA-product_bug/az_angle_bug/T087/products/S1-GUNW-A-R-087-tops-20151111_20150912-153952-30522N_28635N-PP-3c52-v2_0_2.nc" -l 'azimuthAngle' -d download -w ARIAprodrun_testbugfix_T087 -of ISCE -verbose```
Screen Shot 2020-05-30 at 4 06 40 PM

@dbekaert @rzinke, please let me know if you have any suggestions/comments on this

**Comment by dbekaert** [comment]
@sssangha I assume the last profile was swapped NS versus SN, explaining the change in slope?

**Comment by sssangha** [comment]
@dbekaert no, I realized that there was a bug in the code in regards to issue (3) in my previous message. Specifically, the values I wished to purge were not being ignored. Now, I have confirmed that a plane is only fit after truncating the array as intended. See below and note that now the trendline is fixed to the slope corresponding to the majority of pixels:
Screen Shot 2020-05-30 at 10 59 46 PM

**Comment by dbekaert** [comment]
@sssangha this looks good to me. Tagging the team for doing some test and see if this does not introduce any issues.

@team going forward this should be resolved with the PR added in ISCE by @sssangha. The approach above is suggested to handle potential issues in the existing products.

**Comment by sssangha** [comment]
Instead of fitting line along entire azimuth/range, lines are fit in moving window to better isolate more minor artifacts that may be missed in longer scenes.

Furthermore, in previous commits very high thresholds were set in an attempt to capture such minor artifacts (e.g R^2 of 0.998 and standard error on order of 0.0001), however with moving window implementation these thresholds can be relaxed. As a result, in my tests across three tracks false positives have been completely eliminated.

**Comment by dbekaert** [comment]
Thanks @sssangha. @rzinke @ehavazli @bbuzz31 can you give this another attempt. Can you identify cases of false corrections introducing errors?

**Comment by ehavazli** [comment]
looks good to me. produced results are corrected now.
```
ariaDownload.py -i 20151111_20150912 -t 87
ariaExtract.py -f "products/*.nc" -l azimuthAngle -d Download
gdal_translate -of png -scale azimuthAngle/20151111_20150912 azimuthAngle_all.png
```
initial results:
image


new image:

azimuthAngle_all

ehavazli (commented):
it worked for the simplified case summarized by @dbekaert over track 87.
seems to do what it needs to do. I will approve it if it works for everyone else's case tests.

sssangha:
@ehavazli let me first just implement a few minor changes for the ALOS-2 standard product. I will let you know when it's ready

sssangha:
@ehavazli ok now I figured out the source of another bug. In cases where cutlines aren't specified and the imposed bounds supersede the metadata layer extent, then columns/rows that entirely consist of nodata values impose a steep gradient when the metadata layer is interpolated/intersected with the DEM:
Screen Shot 2020-06-12 at 12 13 37 AM

In order to circumvent this, the metadata layer is no longer cropped to the dem bounds before hand, the metadata layer in its native grid is passed through this stage:
Screen Shot 2020-06-12 at 12 13 42 AM

Furthermore, an additional subfunction in the metadata quality control class is added to clean up some redundant routines.

dbekaert (approved):

PR #171, opened 2020-06-04
Rz misclosure

rzinke:


dbekaert:
@rzinke do not see any issues.

Once item i would recommend to do in future is to come up with some unit tests to capture bugs in code changes. The "unit tests" we have for other programs are not really code checks, but program checks. There is an example for the 2-stage stitching on the unwrap branch which is a more of synthetic test where we know what goes in and needs to come out

rzinke:
Thanks, I'll take a look. Should I do the unit checks before you merge the pull request? I tested the code in both the command line, and in the ariaMisclosure_tutorial.ipynb and it works for both of those.

dbekaert:
let merge it for now and them we can enhance during a next PR.

dbekaert (approved):

PR #174, opened 2020-06-09
Bb l cmask

bbuzzanga:
_Continuation of https://github.com/aria-tools/ARIA-tools/pull/169_

Create a mask using the [NLCD dataset](https://www.mrlc.gov/data?f%5B0%5D=category%3Aland%20cover&f%5B1%5D=year%3A2016) by specifying `-m 'nlcd'` e.g.:
-  `ariaExtract.py -f ./products/*.nc -m 'nlcd'`
- `ariaTSsetup.py -f ./products/*.nc -m 'nlcd'`

You can then run `prep_aria.py --water-mask path2mask/NLCD_crop.msk` adjusting path2mask as necessary. There is also a .png of the mask stored in path2mask.


dbekaert (approved):

dbekaert:
@bbuzz31 Although the code passed CircleCI.
The exit code 0 is likely added because of a try except?

dbekaert (commented):
@bbuzz31 

- I noticed there are no changes to the parser for the programs to take benefit from this change.
- Did you also want to add a command line capability for the masking option?

bbuzzanga:
I think the exit code might always trigger? The files appear to be correctly created based on the outputs and size. Yes the parsers aren't changed; the functions are still only called internally. This functionality will be added eventually (and for the DEM too I believe).

tools/ARIAtools/ARIAProduct.py
```diff
@@ -322,7 +322,7 @@ def __continuous_time__(self):
                 num_dups.append(i[0])
         # Delete duplicate products
         self.products=list(self.products for self.products,_ in itertools.groupby(self.products))
-        if num_dups!=0:
+        if len(num_dups) > 0:
```
dbekaert: @sssangha Can you confirm this change.

Askaholic: Idiomatic python would be:
```suggestion
        if num_dups:
```

sssangha:
@bbuzz31 @dbekaert before I merge this PR I wanted input on my proposed changes. I added support for user-specified output formats, whereas before the output was forced to be in ENVI format. However data type UInt16 is not supported by some other output formats, so I instead specified date type Byte (as with the default downloaded mask). 

Is this appropriate? I.e., is there a need to have it in UInt16 format?

dbekaert:
> Is this appropriate? I.e., is there a need to have it in UInt16 format?

Would be appropriate to be consistent.

PR #176, opened 2020-06-20
Fixed bug in average raster generation and reporting failed IFGs

sssangha:
Average rasters through productPlot (coherence) and mask_util (amplitude, when amplitude threshold is specified) were inaccurately being generated. Specifically, ```gdalWarp``` was called on a VRT containing relevant product layers across all pairs, which just results in the first raster being passed. Now each IFG is iterated on in a loop, and a nan-mean is taken after the end of the for-loop. I.e. if at least one IFG has nodata for a given pixel, no value is reported in the final averaged raster.

Since this is executed multiple times through the code depending on the argument, a separate function ```rasterAverage``` is added under ```vrtmanager.py``` to simplify the code. 

Furthermore, now if specified, the average amplitude raster is used to mask the NLCD mask. Before this argument was ignored if a user specified a download of the NLCD mask.

Finally, a minor bug was resolved under ```ARIAproducts.py``` to ensure the appropriate number of total interferograms is reported in the print statements.


dbekaert (approved):
Lets address circleCI separate

Issue #172, opened 2020-06-08
[BUG] AOIassist script


dbekaert:
**Describe the bug**
Error for running AOIassist script.

**To Reproduce**
1. Download 
```
https://api.daac.asf.alaska.edu/services/search/param?intersectsWith=POLYGON((-96.5828%2064.1841,-80.2435%2064.1841,-80.2435%2071.1638,-96.5828%2071.1638,-96.5828%2064.1841))&platform=SENTINEL-1&relativeOrbit=77&processinglevel=SLC&beamSwath=IW&maxResults=5000&output=CSV
```

2. Command used
```
ariaAOIassist.py -f Boothia_Peninsula\ .csv
```
3. Error Output
```
IndexError: list index out of range
(ARIA-tools_latest) MT-101939:NISAR_ST dbekaert$ ariaAOIassist.py -f Boothia_Peninsula\ .csv 
Generating outputs for track: 77
Traceback (most recent call last):
  File "/Users/dbekaert/Software/python/miniconda37/envs/ARIA-tools_latest/bin/ariaAOIassist.py", line 720, in <module>
    track_metadata.plotFrameCenters()
  File "/Users/dbekaert/Software/python/miniconda37/envs/ARIA-tools_latest/bin/ariaAOIassist.py", line 404, in plotFrameCenters
    self.ax.set_xlim([dates[0]-timedelta(days=6),dates[-1]+timedelta(days=6)])
```

dbekaert:
@rzinke running in some other errors when enabling the approximate AOI option.
AttributeError: 'GeometryCollection' object has no attribute 'exterior'

rzinke:
Hi @dbekaert, I tracked down the source of the problem. It is not a bug, but stems from the fact that not all the acquisition epochs cover the same area. Notice how the first two dates of acquisition do not share a common extent, therefore, the program does not know what AOI you want to construct.

I made some modifications so that the routines now automatically exclude epochs for which the spatial extent is not complete. I also added a few checks, warnings, and error messages for cases in which things get wonky. I'll push the changes now, and if they look good I'll update the ARIA-tools-docs.

dbekaert:
@rzinke PR merged

PR #179, opened 2020-06-22
Expanded latitude bounds for ariaAOIassist

rzinke:
I was not able to replicate this bug exactly:
https://github.com/aria-tools/ARIA-tools/issues/172
But I was able to make the file and command work with this new fix.


tools/bin/ariaAOIassist.py
```diff
@@ -32,8 +32,8 @@ def createParser():
             help='Specify directory to deposit all outputs. Default is local directory where script is launched.')
     parser.add_argument('-t', '--tracks', dest='tracks', type=str, default='all',
             help='Include only specified track number in results. Can be multiple, separated by spaces. Default : All')
-    parser.add_argument('-l', '--lat_bounds', dest='latBounds', type=str, default='-60 60',
-            help='Specify a search for only frames that fall within these lat bounds.')
+    parser.add_argument('-l', '--lat_bounds', dest='latBounds', type=str, default='-90 90',
+            help='Specify a search for only frames that fall within these lat bounds. Default : -90 90')
```
dbekaert: @rzinke might be better to make the change at the class init too L74-75?

dbekaert:
@rzinke thanks for fixing

Issue #185, opened 2020-07-09
Add compatibility to GMT grid files


EJFielding:
**Describe the bug**
Output files have unequal spacing and inconsistent geotransform information.

**To Reproduce**
1. Download 
`wget https://aria-share.jpl.nasa.gov/20200623_Oaxaca_Mexico_EQ/Displacements/S1-GUNW_COSEISMIC-A-R-107-tops-20200625_20200619-003043-16794N_15045N-RR-35f2-v2_0_3.nc`
2. Command used
```
ariaExtract.py -f "products/S1-GUNW-A-R-107-tops-20200625_20200619-003043-16794N_15045N-RR-35f2-v2_0_3.nc" -l all -d download -w products -m download --outputFormat ENVI
gdal_calc.py -A products/unwrappedPhase/20200625_20200619.vrt -B products/connectedComponents/20200625_20200619.vrt --outfile results/S1_A107_20200625_20200619_disp_cm.tif --calc='((A*(B>0))/(4*pi)*5.546576)' --format=GTiff --NoDataValue=0 --overwrite
gdal_translate -of GMT S1_A107_20200625_20200619_disp_cm.tif S1_A107_20200625_20200619_disp_cm.grd
gmt grdlandmask -Glandmask.grd -RS1_A107_20200625_20200619_disp_cm.grd
gmt grdinfo landmask.grd
```
3. Error Output
```
grdinfo [ERROR]: (x_max-x_min) must equal (NX + eps) * x_inc), where NX is an integer and |eps| <= 0.0001.
grdinfo (gmtlib_read_grd_info): Use grdedit -A on your grid file to make region and increments compatible [landmask.grd]
landmask.grd: Title: z
landmask.grd: Command: grdlandmask -Glandmask.grd -RS1_A107_20200625_20200619_disp_cm.grd
landmask.grd: Remark: Derived from the low resolution shorelines
landmask.grd: Pixel node registration used [Geographic grid]
landmask.grd: Grid file format: nf = GMT netCDF format (32-bit float), CF-1.7
landmask.grd: x_min: -96.5451065365 x_max: -93.9876553945 x_inc: 0.00083331741351 (3 sec) name: longitude n_columns: 3069
landmask.grd: y_min: 15.046045981 y_max: 16.7804535652 y_inc: 0.000833449103383 name: latitude n_rows: 2081
landmask.grd: z_min: 0 z_max: 1 name: z
landmask.grd: scale_factor: 1 add_offset: 0
[...]

PR #186, opened 2020-07-10
Fixed mask nodata value to 0

sssangha:
For certain output formats (e.g. GSBG see issue #185) by default the nodata value for the mask file is set to Inf/-Inf. Thus, when the raster is read as a numpy array, nodata values are passed as Inf/-Inf through the program. This is overridden by defining the nodata value through gdal as 0.

### DIFF
--- a/tools/ARIAtools/mask_util.py
+++ b/tools/ARIAtools/mask_util.py
@@ -57,7 +57,7 @@ def prep_mask(product_dict, maskfilename, bbox_file, prods_TOTbbox, proj, amp_th
         gdal.Rasterize(os.path.join(workdir,'watermask_uncropped.msk'), os.path.join(workdir,'watermsk_shorelines.vrt'), options=gdal.RasterizeOptions(format=outputFormat, outputBounds=bounds, outputType=gdal.GDT_Byte, width=arrshape[1], height=arrshape[0], burnValues=[1], layers='merged'))
         gdal.Translate(os.path.join(workdir,'watermask_uncropped.msk.vrt'), os.path.join(workdir,'watermask_uncropped.msk'), options=gdal.TranslateOptions(format="VRT"))
         # save cropped mask
-        gdal.Warp(maskfilename, os.path.join(workdir,'watermask_uncropped.msk.vrt'), options=gdal.WarpOptions(format=outputFormat, outputBounds=bounds, outputType=gdal.GDT_Byte, width=arrshape[1], height=arrshape[0], multithread=True, options=['NUM_THREADS=%s'%(num_threads)]))
+        gdal.Warp(maskfilename, os.path.join(workdir,'watermask_uncropped.msk.vrt'), options=gdal.WarpOptions(format=outputFormat, outputBounds=bounds, dstNodata=0, outputType=gdal.GDT_Byte, width=arrshape[1], height=arrshape[0], multithread=True, options=['NUM_THREADS=%s'%(num_threads)]))
         update_file=gdal.Open(maskfilename,gdal.GA_Update)
         update_file.SetProjection(proj)
         update_file.GetRasterBand(1).SetNoDataValue(0.) ; del update_file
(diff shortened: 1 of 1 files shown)


dbekaert:
This means that you are combining no-data with water.

EJFielding:
In GMT, they set nodata pixels to NaN.
@dbekaert I am not sure it is important to distinguish between no-data outside of interferogram from water mask. We don't want to use either one for further analysis.

sssangha:
@EJFielding so would it be more appropriate to just redefine the loaded numpy alone. Specifically, see line 119:
```
        mask=gdal.Warp('', maskfilename, options=gdal.WarpOptions(format="MEM", cutlineDSName=prods_TOTbbox, outputBounds=bounds, dstNodata=0, width=arrshape[1], height=arrshape[0], multithread=True, options=['NUM_THREADS=%s'%(num_threads)]))
```

I.e., in your experience is it appropriate to define GMT files with nodata values of 0.

EJFielding:
@sssangha Yes, not a problem if nodata value is zero. I can use a simple `grdmath` command to set the zero pixels to NaN. I do that all the time with ISCE output.

dbekaert:
> In GMT, they set nodata pixels to NaN.
> @dbekaert I am not sure it is important to distinguish between no-data outside of interferogram from water mask. We don't want to use either one for further analysis.

You cannot distinguish in the case for example the water mask is incomplete versus land.
for example in the occurrence of missing tiles.

dbekaert:
@sssangha we do set the no-data meta-data correctly. 
Is this not an item to be fixed by the GDAL developers that they take into account the no-data for the GSBG option?

EJFielding:
@sssangha @dbekaert I am not a GDAL expert, but I remember something about having to set or get the nodata value for dataset objects.

dbekaert:
@EJFielding we internally tag all the no-data value. If you extract the file as a vrt or an envi file and do gdal info on the file it will report you the no-data value. Changing the output type to GSBG and seeing these artifact tells me that they are not considering the meta-data attributes.

Perhaps let me ask it different:

if you extract the file as a envi file or a vrt. If you do a gdalinfo on the file is the no-data value correctly handled?
If yes; what happens if you now translate this now to an GSBG file. 
I understand that now you are experiencing issues with no-data?
If yes, then i would say this should be reported to GDAL

EJFielding:
@dbekaert The unwrapped phase output in GSBG format with the landmask applied had the ocean set to two different values (-Inf and +Inf) in some strange pattern. There was also an error reported, so I understand that @sssangha was trying to avoid the error report.

EJFielding:
@sssangha I tried your PR version, and it no longer reports an error for GSBG output. I was able to open the new unwrappedPhase output file in QGIS, but I had to add zero as an "additional nodata value". Then it looks the same as the ENVI output file. It seems that the GSBG file has +Inf as the specified "nodata" value as QGIS is picking up that. I can use this file, but I think it would be better to have the "nodata" pixels set to the "nodata" value in the GSBG header. This probably means that the mask file can't be loaded from GSBG format and then used without first converting the "nodata" pixels to zero before using it for multiplication, which is a few extra steps.

sssangha:
@EJFielding after some investigation, I realized that while the rasters are loaded as expected through numpy, there is no way to overwrite the default nodata value for the ```GSBG``` format. Thus, it looks like a bug on the gdal end, so I have posted an issue ticket on their github page: https://github.com/OSGeo/gdal/issues/2759

Unfortunately, even once this is resolved there will likely be a significant lag before the gdal conda channel is updated for a potential bug-fix.

bbuzzanga:
There seems to still be a problem where the inf values aren't getting replaced. I get the following incorrect mask (left) when using GSBG output and reading the resulting numpy array directly (the correct result is on the right, not using the GSBG option). This directly affects later processing, e.g. coherence computations.
Screen Shot 2020-07-14 at 19 48 06

EJFielding:
@sssangha @bbuzz31 I can understand that it probably requires quite a bit more code, but it seems better not to try to change the `nodata` value for the GSBG output (or other formats). This means that every read of files in those formats will require a call to `srcband.GetNoDataValue()` for each band and code to convert those pixels to whatever value is appropriate for the further processing steps like applying a mask. I need to add this kind of code to the earthquake modeling software I am using because that code is not reading the output of ariaExtract correctly. That is what started my original #185 bug report.

dbekaert:
At this point i do not think its worth to pursue this in our code. @sssangha rather print out a warning message to the user in case this GDAL output option is selected. Per discussion on the GDAL forum the design of overwriting this is done by reason. All other export options work fine and with no-date properly being propagated and captured.

@EJFielding, we will see if we can provide an alternative post-processing option to go into GMT.

Issue #188, opened 2020-07-13
[BUG] ariaTSsetup.py with -mo option fails


EJFielding:
**Describe the bug**

I added the `-mo 15000` option to my `ariaTSsetup.py` run to select only the pairs with minimumOverlap of greater than 15000 km^2. It worked for one bbox, but not for another.

**To Reproduce**
1. Download 
`ariaDownload.py --bbox "26 28 83 86" --track 19 -o download`
2. Command used
```
# unit of minimumOverlap are km^2
ariaTSsetup.py -f 'products/*.nc' --bbox "26 28 83 86" -d download --mask download -w bbox26_28 -mo 15000
```
3. Error Output
```
***Time-series Preparation Function:***
Shapefile bbox26_28/user_bbox.json created for input user bounds.
Multi-core version
Multi-core version failed, will try single for loop
285 out of 470 GUNW products rejected for not meeting user's bbox spatial criteria
Group GUNW products into spatiotemporally continuous interferograms.
All (163) interferograms are spatially continuous.
Thread count specified for gdal multiprocessing = 2
Traceback (most recent call last):
  File "/Users/fielding/anaconda3/envs/aria-tools/bin/ariaTSsetup.py", line 15, in <module>
    main(inps)
  File "/Users/fielding/anaconda3/envs/ARIA-tools/lib/python3.8/site-packages/ARIAtools/tsSetup.py", line 254, in main
    standardproduct_info.products[0], standardproduct_info.products[1], standardproduct_info.bbox_file, prods_TOTbbox, prods_TOTbbox_metadatalyr, arrshape, proj = merged_productbbox(standardproduct_info.products[0], standardproduct_info.products[1], os.path.join(inps.workdir,'productBoundingBox'), standardproduct_info.bbox_file, inps.croptounion, num_threads=inps.num_threads, minimumOverlap=inps.minimumOverlap, verbose=inps.verbose)
  File "/Users/fielding/anaconda3/envs/ARIA-tools/lib/python3.8/site-packages/ARIAtools/extractProduct.py", line 226, in merged_productbbox
    overlap_area=shapefile_area(prods_bbox)
  File "/Users/fielding/anaconda3/envs/ARIA-tools/lib/python3.8/site-packages/ARIAtools/shapefile_util.py", line 77, in shapefile_area
    lon, lat=file_bbox.exterior.coords.xy
AttributeError: 'MultiPolygon' object has no attribute 'exterior'
```
**Expected behavior**

[...]

PR #190, opened 2020-07-15
Adjust bbox to IFG grid, handle multipolygons

sssangha:
Addressed the following two issues:
1. Under the ```merged_productbbox``` function of the ```extractProduct.py``` script, bounding box shapefiles and bounds are now adjusted to ensure they are integer multiples of the standard interferometric grid increment. This is done by passing the 'target aligned pixels' option to the first call to gdalWarp in order to access the master bounds. These bounds are then inherited by all future warping calls. Addresses discussion in Issue #185 
2. Handle area calculations for fringe cases where intersection of products and bounding boxes produces multipolygons (see Issue #188 for illustrations). Changes made to```shapefile_area``` function under the ```shapefile_util.py``` script.

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -166,6 +166,7 @@ def merged_productbbox(metadata_dict, product_dict, workdir='./', bbox_file=None
 
     # Import functions
     from ARIAtools.shapefile_util import save_shapefile, shapefile_area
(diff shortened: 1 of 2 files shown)


dbekaert:
@sssangha please disentangle the PR. The area change is trivial the sampling change i would like to see vetted prior to merging.

PR #191, opened 2020-07-15
Handle multipolygon intersections

sssangha:
Handle area calculations for fringe cases where intersection of products and bounding boxes produces multipolygons (see Issue #188 for illustrations). Changes made to ```shapefile_area``` function under the ```shapefile_util.py``` script.


dbekaert (approved):
@sssangha looks good

PR #192, opened 2020-07-15
Adjust output bounds to be multiple of increment

sssangha:
Under the ```merged_productbbox``` function of the ```extractProduct.py``` script, bounding box shapefiles and bounds are now adjusted to ensure they are integer multiples of the standard interferometric grid increment. This is done by passing the 'target aligned pixels' option to the first call to ```gdalWarp``` in order to access the master bounds. These bounds are then inherited by all future warping calls. Addresses discussion in Issue #185 

I am concerned in cases where users expect outputs to be strictly defined by their imposed output bounds for their own post-processing steps, so some thoughts/advice would be welcome.

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -166,6 +166,7 @@ def merged_productbbox(metadata_dict, product_dict, workdir='./', bbox_file=None
 
     # Import functions
     from ARIAtools.shapefile_util import save_shapefile, shapefile_area
+    from shapely.geometry import Polygon
 
     # If specified workdir doesn't exist, create it
     if not os.path.exists(workdir):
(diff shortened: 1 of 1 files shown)


dbekaert:
@sssangha: we should jump to the nearest bounds of a user requested area while maintaining the original posting. I do not foresee a problem there.

EJFielding:
@sssangha @dbekaert Yes, I think it is reasonable to jump to nearest multiple of the 3-arcsecond ARIA product spacing. I hope this fixes issue #185 that I raised earlier about the `ariaExtract` products being resampled from the original product grid.

ehavazli (approved):

ehavazli:
I checked the pixel size details and ran the commands from #185 . seems to return the correct values and successfully produce expected results.

EJFielding:
I tried this out with my data, now that it is merged. I am not very proficient with trying out the PR branches. @sssangha The new version works great and produces output files with the original  3-arcsecond spacing and consistent region limits. Thanks!

There is one small related question about the type of pixel registration used in the ARIA-tools processing. In GMT, they explicitly record whether the pixel locations are specified at the center of the pixel (gridline registration) or the top left corner of the pixel (pixel registration). The original SRTM DEM is gridline registration, so it has 1201 points across a one-degree tile of 3-arcseond data because one pixel is centered on the first degree and the last one is centered on the next degree line. I believe that GDAL always uses pixel registration, which means there is a shift of half a pixel from the gridline registration and usually there are 1200 pixels per degree. I understand the original ARIA standard product is geocoded with the SRTM so it should have the gridline registration. When I use GMT `grdconvert` directly on the ARIA NetCDF file, the output has the gridline registration. 

The question is whether the `ariaExtract` internal calculations and output are done with the pixel registration or the gridline registration. It looks like the output is pixel registration. Does this mean it is internally interpolating all the pixels to account for the half-pixel shift? In most cases, we don't need to worry about half a pixel location accuracy but it would be good to document what you are doing.

PR #196, opened 2020-07-23
Automatic figure resize for productPlot.

rzinke:
If ariaPlot.py -plottracks is run with more than a dozen or so pairs, the pair labels will overlap and become unreadable. This fix automatically scales the figure size so that the pair labels do not overlap.

### DIFF
--- a/tools/ARIAtools/productPlot.py
+++ b/tools/ARIAtools/productPlot.py
@@ -249,8 +249,11 @@ class plot_class:
             Make plot of track extents vs bounding bbox/common track extent.
         '''
 
-        ax=self.plt.figure().add_subplot(111)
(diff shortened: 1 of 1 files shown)


EJFielding:
This new version is convenient if the user has Adobe Illustrator or some other convenient way to view a very wide `.eps` plot. Here is an example from Nepal where I loaded the `lat_extents.eps` into Illustrator and then exported to PNG:
lat_extents-wide

In some other cases, people might want the old-style compressed plot to get a quick overview of the coverage, like this plot from the old version:
lat_extents-new

@rzinke Would it be possible to add an option for wide-and-readable or narrow-overview plots?

PR #213, opened 2020-08-19
Pass gdalWarp update to 1.1.1

sssangha:


dbekaert (changes requested):
@sssangha can you first Pull the release-v1.1.1 branch and then copy your fix into it for GDAL. This PR has a mixture and also includes the logging parts.

Issue #194, opened 2020-07-21
[BUG] meta-data verification 


rzinke:
**Describe the bug**
Not quite sure where to start with this error message. I am running ariaTSsetup.py on the new Tibet track 41 data. Lately, I moved all the old v 2.0.2 data to a separate folder and downloaded all the new, v 2.0.3 data to the folder Track041/AllProducts. When I run TSsetup, I now get an error message in the finalize_metadata section (see below).

**To Reproduce**
1. Download 
ariaDownload.py -t 41 -w Track041/AllProducts
2. Command used
ariaTSsetup.py -f '/u/sar-r2/rzinke/Tibet/Track041/AllProducts/*.nc' -w ML3v3 -d download -m download -nt 8 -ml 3
3. Error Output
Traceback (most recent call last):
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/bin/ariaTSsetup.py", line 15, in <module>
    main(inps)
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.6/site-packages/ARIAtools/tsSetup.py", line 275, in main
    export_products([dict(zip([k for k in set(k for d in standardproduct_info.products[1] for k in d)], [[item for sublist in [list(set(d[k])) for d in standardproduct_info.products[1] if k in d] for item in sublist] for k in set(k for d in standardproduct_info.products[1] for k in d)]))], standardproduct_info.bbox_file, prods_TOTbbox, layers, dem=demfile, lat=Latitude, lon=Longitude, mask=inps.mask, outDir=inps.workdir, outputFormat=inps.outputFormat, stitchMethodType='overlap', verbose=inps.verbose, num_threads=inps.num_threads, multilooking=inps.multilooking)
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.6/site-packages/ARIAtools/extractProduct.py", line 343, in export_products
    finalize_metadata(outname, bounds, dem_bounds, prods_TOTbbox, dem, lat, lon, mask, outputFormat, verbose=verbose)
  File "/u/sar-r0/rzinke/python/miniconda3/envs/ARIA-tools/lib/python3.6/site-packages/ARIAtools/extractProduct.py", line 615, in finalize_metadata
    data_array = metadata_qualitycheck(data_array, os.path.basename(os.path.dirname(outname)), outname, verbose).data_array
[...]

PR #215, opened 2020-08-20
Properly handle nans in metadata layer extraction

sssangha:
Metadata quality control class ```metadata_qualitycheck``` now appropriately handles nodata values.

Where artifacts are present for a given layer in the cube, the general strategy is to remove a best fit plane using the ```lstsq``` module from ```scipy```. Thus it is essential to purge each layer of nodata values before fitting this plane. 

The challenge is to work around the limitations of the ```lstsq``` module in handling masked values, and  also making sure to pass arrays that accurately track the row/col indices of each valid pixel as arguments. 

The original logic scheme was a bit convoluted as all rows and columns containing any masked values (which is assumed to correspond to the edge of stitched boundaries where there are frame shifts) were masked entirely, which simplifies the procedure of recording row/col indices arrays for valid pixels. This becomes a problem for exceptionally large AOIs spanning several frames where nothing would be left if the layers were masked in such a matter (see #194)

Now I have implemented a new streamlined scheme which purges only masked values from the metadata array before initiating a planar fit, while also making sure the expected row/col indices array inputs are successfully purged of masked pixel indices.

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -414,33 +414,30 @@ class metadata_qualitycheck:
         # Run class
         self.__run__()
 
-    def __truncateArray__(self, data_array_band, Xmask, Ymask, direction='col'):
-        #crop out columns with any masked values
-        if direction=='col':
-            Xmask=Xmask.T[np.all(data_array_band.T.mask == False, axis=1)].T
-            Ymask=Ymask.T[np.all(data_array_band.T.mask == False, axis=1)].T
-            data_array_band=data_array_band.T[np.all(data_array_band.T.mask == False, axis=1)].T
-        #crop out rows with any masked values
-        if direction=='row':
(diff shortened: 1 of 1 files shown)


dbekaert:
@rzinke can you try this out and confirm it works?

@sssangha it would be good to start including a few unit test and think about capturing exceptions for the logger?

rzinke:
Great, thanks very much! I’ll test this today.

sssangha:
@dbekaert yeah, the latter should be straightforward. But in regards to unit tests, I don't see what we could setup that isn't captured/vetted by circleCI? Speaking of which, we talked before about maturing our circleCI setup to better articulate errors as is done in RAiDER

dbekaert:
@sssangha i think that is a nice to have, but less urgent that e.g. ability to continue processing, logging etc. Which are all new functionality. The pytest is just a nicer way to show it.

dbekaert (approved):
vetted by @sssangha and @rzinke

PR #219, opened 2020-08-31
Update S1 sceneLength to 35

sssangha:
SceneLength key was originally hardcoded to 27 seconds for S1 products. This value was leveraged to compare adjacent frames and determine whether or not they temporally overlap.

However, upon closer inspection I saw that this value varies between 27-35 seconds. Thus when comparing adjacent frames with a sceneLength of 27 vs 35 seconds, the program would flag a gap even though the products were actually continuous. This issue is resolved by simply setting the ```sceneLength``` key to 35 seconds, which would avoid such positives but still flag legitimate gaps.

As tests, please run the following examples:
1. False positive gap (now circumvented)
ariaDownload.py -t 144 -i 20181117_20181018
ariaExtract.py -f "products/*20181117_20181018*.nc" -v -w falsepositivegap_20181117_20181018

Using python, take a closer look at product "S1-GUNW-D-R-144-tops-20181117_20181018-135957-35740N_33435N-PP-763f-v2_0_2.nc'" to see the upper end sceneLength of 35 seconds:
```
import netCDF4
first='products/S1-GUNW-D-R-144-tops-20181117_20181018-135957-35740N_33435N-PP-763f-v2_0_2.nc'
#first='products/S1-GUNW-D-R-144-tops-20181117_20181018-135928-37232N_35260N-PP-b467-v2_0_2.nc'
rdrmetadata = netCDF4.Dataset(first, keepweakref=True,diskless=True).groups['science'].groups['radarMetaData']
rdrmetadata
KeyboardInterrupt
rdrmetakeys = list(rdrmetadata.variables.keys())
rdrmetakeys
['missionID', 'wavelength', 'centerFrequency', 'productType', 'ISCEversion', 'unwrapMethod', 'DEM', 'ESDthreshold', 'azimuthZeroDopplerStartTime', 'azimuthZeroDopplerEndTime', 'azimuthTimeInterval', 'slantRangeSpacing', 'slantRangeEnd', 'slantRangeStart']
rdrmetadata['azimuthZeroDopplerStartTime'][0]
#'2018-11-17T13:59:15.420122Z'
rdrmetadata['azimuthZeroDopplerEndTime'][0]
#'2018-11-17T13:59:42.366409Z'
```
Uncomment line 3 to take a look at the adjacent product  to see the lower end sceneLength of 27 seconds

2. Legitimate gap (still correctly flagged)
ariaDownload.py -t 144 -i 20160606_20150519
ariaExtract.py -f "products/*20160606_20150519*.nc" -v -w legitimategap_20160606_20150519

### DIFF
--- a/tools/ARIAtools/ARIAProduct.py
+++ b/tools/ARIAtools/ARIAProduct.py
@@ -263,7 +263,7 @@ class ARIA_standardproduct: #Input file(s) and bbox as either list or physical s
                 rdrmetadata_dict['slantRangeStart']=798980.125
                 rdrmetadata_dict['slantRangeEnd']=956307.125
                 #hardcoded key meant to gauge temporal connectivity of scenes (i.e. seconds between start and end)
-                rdrmetadata_dict['sceneLength']=27
+                rdrmetadata_dict['sceneLength']=35
             elif basename.split('-')[0]=='ALOS2':
                 rdrmetadata_dict['missionID']='ALOS-2'
                 rdrmetadata_dict['productType']='UNW GEO IFG'


ehavazli:
@sssangha the approach you adopt seems reasonable for the purposes of ARIA-tools. However, I am curious if it is better to implement more global solution by calculating the actual time difference and passing it through an if loop and confirm it is between 27-35 seconds? I am curious to hear your opinions.
@dbekaert @bbuzz31

dbekaert:
@sssangha might be good in future to remove the dependency on scene length. Originally we had it extracted from the product, which was accurate, now with hard-coding it becomes more tricky. However we could do the following in the future to sort the files:

1) use the reference and secondary dates. You can filter in time based on the 6-day revisit for Sentinel-1; anything less  that 6 day difference is on the same pass, which would allow for a midnight crossing scenario.
2) for spatial sub-setting and finding gaps in space one could use the north south extent from the filename to figure it out,. We should keep the legacy approaches as separate function if we ever want to have

dbekaert:
> @sssangha the approach you adopt seems reasonable for the purposes of ARIA-tools. However, I am curious if it is better to implement more global solution by calculating the actual time difference and passing it through an if loop and confirm it is between 27-35 seconds? I am curious to hear your opinions.
> @dbekaert @bbuzz31

We can do it very accurately if we crack open the product. Although we stepped away from that given that did not allow for virtual data access. Once could asses how consistent this is for the products in the global archive, alternative see the comment i made above.

dbekaert (approved):
@sssangha @ehavazli for now this is OK, but we should think long-term and sensor neutral implementation

EJFielding:
Yes, I hope we eventually have ALOS-2 data products in ARIA.

Issue #232, opened 2020-10-05
[BUG] ARIA stitched products give zero values


rzinke:
I am following up on a bug I originally discovered using MintPy, and reported there: https://github.com/insarlab/MintPy/issues/453

However, indications point to the bug originating in the ARIA-tools workflow. Essentially, the binary image files are stored with all nodata values. I tested this for two Tibet tracks and both turned out this way.

dbekaert:
@rzinke Can you provide additional information such other can replicate your problem?
Also did you re-run this from an existing tun or a clean run?

The png that you showed looks to only provide partial coverage over the region of interest.
The pngs are generated after the data is written to the envi file: https://github.com/aria-tools/ARIA-tools/blob/b787e827a799fecbf45d6aa267eeb95940148d83/tools/ARIAtools/unwrapStitching.py#L300
The files itself are not touched anymore afterwards, as the stack files are merely pointing to these files.

Can you investigate further if this is applicable to all the files; e.g. .coherence, connected components, phase, geometric files etc?

rzinke:
Okay. It appears there are several things going on here:

First, I want to use a practice data set for troubleshooting. I downloaded the standard products:
```
ariaDownload.py -t 19 -s 20190101 -e 20190601 
```
then launched the ARIA TS setup script: 
```
ariaTSsetup.py -f "products/*.nc" -b 19_A_Tibet_AOI.shp -l 3 -d download -m download -nt 8
```

which produces the following error:
```
NETCDF:"/u/sar-r2/rzinke/ARIAfix/T19/products/S1-GUNW-D-R-019-tops-20190526_20190514-001637-39721N_37749N-PP-22b6-v2_0_2.nc":/science/grids/data/connectedComponents is GDAL compatible
Generating: unwrappedPhase - [======================= 98% ====================> ] 20190601_20190508  385s /     7sNETCDF:"/u/sar-r2/rzinke/ARIAfix/T19/products/S1-GUNW-D-R-019-tops-20190601_20190508-001444-44029N_42063N-PP-1f59-v2_0_2.nc":/science/grids/data/unwrappedPhase is GDAL compatible
NETCDF:"/u/sar-r2/rzinke/ARIAfix/T19/products/S1-GUNW-D-R-019-tops-20190601_20190508-001444-44029N_42063N-PP-1f59-v2_0_2.nc":/science/grids/data/connectedComponents is GDAL compatible
[...]

PR #234, opened 2020-10-08
Specified download of compatible GDAL version

sssangha:
GDAL versions > 3.0.4 may generate empty outputs (refer to #232). Until this bug is addressed by GDAL, we will fix the install to point to the most recent working version.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -10,7 +10,7 @@ channels:
 dependencies:
   - python>=3.6
   - cartopy
-  - gdal>=3.0
+  - gdal=3.0.4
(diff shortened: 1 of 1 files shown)


dbekaert:
@sssangha this is very strict. Suggest to use >=3.0 and <=3.0.4; i.e. give a range

dbekaert (changes requested):
see comment

dbekaert:
@ehavazli is doing a test.

dbekaert:
This has been merged via #235

Issue #238, opened 2020-10-31
[BUG] bring own DEM and water mask error message "_uncropped"


dbekaert:
**Describe the bug**
when bringing own DEM it looks for file*_uncropped.vrt; which does not exist when bringing own DEM.

**To Reproduce**
1. Download dem from dem.py .e.g ISCE
2. run ariaExtract.py for a geometry layer using the downloaded dem.

EJFielding:
Not only does `ariaExtract` not find the DEM that is provided by the user, it erases all the DEM files when it fails.

dbekaert:
Good point @EJFielding; this would not be desired. It is likely related to the "uncropped"; which was meant to be the untouched file. We will look into this.

EJFielding:
I find a similar bug when I try to specify the mask file to avoid downloading it again.
```
ariaExtract.py -f S1-GUNW-D-R-131-tops-20201030_20201024-160721-38516N_36802N-RR-d813-v2_0_3.nc -l all -m mask/watermask.msk.vrt
***Extract Product Function:***
All (1) GUNW products meet spatial bbox criteria.
Group GUNW products into spatiotemporally continuous interferograms.
All (1) interferograms are spatially continuous.
All layers are to be extracted, pass all keys.
Thread count specified for gdal multiprocessing = 2
Traceback (most recent call last):
  File "/Users/fielding/anaconda3/envs/ARIA-tools/lib/python3.8/site-packages/ARIAtools/mask_util.py", line 111, in prep_mask
    gdal.Warp(maskfilename, os.path.join(workdir,os.path.basename(maskfilename).split('.')[0]+'_uncropped.msk.vrt'), options=gdal.WarpOptions(format=outputFormat, cutlineDSName=prods_TOTbbox, outputBounds=bounds, width=arrshape[1], height=arrshape[0], multithread=True, options=['NUM_THREADS=%s'%(num_threads)]))
  File "/Users/fielding/anaconda3/envs/ARIA-tools/lib/python3.8/site-packages/osgeo/gdal.py", line 613, in Warp
    srcDSTab = [Open(srcDSOrSrcDSTab)]
  File "/Users/fielding/anaconda3/envs/ARIA-tools/lib/python3.8/site-packages/osgeo/gdal.py", line 3308, in Open
    return _gdal.Open(*args)
RuntimeError: ./mask/watermask_uncropped.msk.vrt: No such file or directory

During handling of the above exception, another exception occurred:

[...]

Issue #237, opened 2020-10-23
Error on downloading DEM files


nabolfat:
Hi,
I am trying to use ariaTSsetup running the following command:
ariaTSsetup.py -b '35.2 36.2 -118 -117' -f 'products/*.nc' --mask Download -w 'products_ariaoutputs'

The DEM directory is made and is empty, and then I will get the following error message:
"RuntimeError: products_A_t64_ariaoutputs/DEM/SRTM_3arcsec_uncropped.dem.vrt: No such file or directory"
 
The DEM files should be downloaded automatically. I could not figure out why I am getting the error message.
I appreciate it if you can help me with this.

Thanks,
Niloufar

dbekaert:
Changes on opentopo for hosting the DEM. We are looking for a work-around.

rzinke:
Hello, I am running into the same issue. Have there been any recent updates?

dbekaert:
@rzinke, the solution is implemented here: https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/dem.py; would need porting to ARIA-tools in the same way.

yuankailiu:
@dbekaert, @rzinke, is there a way currently to use DEM which is pre-downloaded ourselves before hand (i.e., from [GMRT.org](https://www.gmrt.org/GMRTMapTool/))? 

I also tryied to do the same way as the implementation in [RAiDER](https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/dem.py).

Is it this line in ARIA-tools' extractProduct.py needs to be modified for porting?
`_world_dem = '/vsicurl/https://cloud.sdsc.edu/v1/AUTH_opentopography/Raster/SRTM_GL1_Ellip/SRTM_GL1_Ellip_srtm.vrt'`

I replace it by the following:
`    _world_dem = "https://portal.opentopography.org/API/globaldem?demtype=SRTMGL1_E&south={}&north={}&west={}&east={}&outputFormat=GTiff"`

Do I miss something that it still doesn't work?

dbekaert:
## fix bring own DEM
@liuykai wrt own DEM, just it was always envisioned to be supported, but needs a bug fix see #238.

## fix download internally in ARIA-tools
Wrt porting over the fix from RAiDER here:
1. update the filepath
[...]

PR #239, opened 2020-10-31
fix dl issue in regards to open topo change

bbuzzanga:
addresses #237, #238

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -11,6 +11,7 @@ import numpy as np
 import glob
 from osgeo import gdal
 import logging
+import requests
 from ARIAtools.logger import logger
 
 from ARIAtools.shapefile_util import open_shapefile
@@ -22,6 +23,8 @@ gdal.UseExceptions()
 gdal.PushErrorHandler('CPLQuietErrorHandler')
 
 log = logging.getLogger(__name__)
+
+_world_dem = "https://portal.opentopography.org/API/globaldem?demtype=SRTMGL1_E&west={}&south={}&east={}&north={}&outputFormat=GTiff"
 
 def createParser():
     '''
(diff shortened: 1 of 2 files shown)


dbekaert:
Thanks @bbuzz31. @sssangha i looked through this but do not directly find the reason why circle CI is  failing. Could you have  another look. Recall you had a similar issue in the past?

dbekaert:
Looks good to me @bbuzz31, some minor codacy issues.
Thanks for porting it over.

dbekaert (approved):

dbekaert:
@bbuzz31 to clarify it solved bring own DEM - not the own watermask issue; and the download of DEM.

bbuzzanga:
@dbekaert that was correct but this latest push stops the removal of the watermask. Codacy doesn't like the checks to make sure a user defined DEM is gdal compatible; I'm not sure how to make it happy.

EJFielding:
I tried the PR and this works for me to use a DEM downloaded with ISCE2 `dem.py` and it also reused my previously downloaded watermask without deleting it. Thanks @bbuzz31 !

bbuzzanga:
Great @EJFielding!

liulinan:
Hi @EJFielding,  I am running into the same issue that I can't do the DEM download. You mentioned you tried the PR and solved this issue. Could you please provide some more details, like what the PR represent? Sorry that I'm a new beginner. Thanks!

EJFielding:
Hi @liulinan, here on GitHub "PR" stands for "pull request". When bbuzz31 made some changes to fix the problem, he sent a pull request, which has number #239 , to ARIA-tools with the changes. Those changes have now been merged into the main ARIA-tools version, so you don't need to use the PR version. I checked the PR before it was merged to verify that it fixed my problem. You can see that GitHub shows above that my issue #238 was closed after I verified that the changes fixed the issue that I reported.

liulinan:
Thanks @EJFielding . When I tried to download DEMs, I got an error "RuntimeError: ./DEM/SRTM_3arcsec_uncropped.dem.vrt: No such file or directory". I'm kinda lost after going throgh so many fixed solutions. Could you please tell me which is the final solution I can use to solve this issue? Thanks again!

EJFielding:
The changes in this PR #239 were merged into the main version, so you only need to update your installation to the latest main version by running `git pull` in your ARIA-tools directory and then re-installing with `python setup.py install`.

liulinan:
That makes sense to me. Many thanks @EJFielding

Issue #244, opened 2020-11-25
[BUG] ampStack.vrt file missing after "ariaTSsetup.py --layers all"


yunjunz:
**Describe the bug**

From the script help message, I would expect `ariaTSsetup.py --layers all` will generate amplitude / bPerpendicular / bParallel in a similar fashion as the coherence dataset. However, 

+ `amplitude` folder is successfully generated with files inside, but a `stack/ampStack.vrt` file is missing.
+ `bParallel` folder is successfully generated with files inside, but a `stack/bParStack.vrt` file is missing.
+ both `bPerpendicular` folder and `stack/bPerpStack.vrt` are missing.

**To Reproduce**
1. Download: `ariaDownload.py -b '37.25 38.1 -122.6 -121.75' --track 42 --end 20200315`
2. Command used: `ariaTSsetup.py -f 'products/*.nc' -b '37.25 38.1 -122.6 -121.75' -l all --mask download`

**Expected behavior**
It will be great if all of them can be generated. Thank you!

**Desktop (please complete the following information):**
 - ARIA-tools git tag: [e.g. git show --summary]
```
commit 86f05a5440acb612b6a396428bba5afd49eaf4ce
Merge: d47e12a 6ea9300
Author: ehavazli <emre.havazli@jpl.nasa.gov>
Date:   Tue Nov 10 19:50:41 2020 -0800

    Merge pull request #243 from alexdunncs/dev
    
    Fixed/removed deprecated ogr and gdalconst imports
```
 - OS: `Red Hat Version 7.8`

**Additional context**

yunjunz:
The main issue of `ampStack.vrt` file has been fixed by #245. Thank you @sssangha. The `bParallel`/`bPerpendicular` folder does not seem important to me. Please feel free to close this issue if you like.

PR #245, opened 2020-12-15
Generate amplitude stack

sssangha:
Virtual amplitude stack now generated, if user specifies extraction of amplitude layer (or all layers) when running ```ariaTSsetup.py```

@yunjunz this is of relevance to a mintpy PR (https://github.com/insarlab/MintPy/pull/481).

### DIFF
--- a/tools/ARIAtools/tsSetup.py
+++ b/tools/ARIAtools/tsSetup.py
@@ -125,7 +125,7 @@ def generateStack(aria_prod,inputFiles,outputFileName,workdir='./', refDlist=Non
         domainName = 'Coherence'
         cohList = glob.glob(os.path.join(workdir,'coherence','[0-9]*[0-9].vrt'))
         dataType = "Float32"
-        print('Number of coherence discovered: ', len(cohList))
+        print('Number of coherence files discovered: ', len(cohList))
         Dlist = sorted(cohList)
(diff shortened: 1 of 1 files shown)


tools/ARIAtools/tsSetup.py
```diff
@@ -301,6 +307,8 @@ def main(inps=None):
         if layers!=[]:
             print('\nExtracting optional, user-specified layers %s for each interferogram pair'%(layers))
             export_products(standardproduct_info.products[1], standardproduct_info.bbox_file, prods_TOTbbox, layers, dem=demfile, lat=Latitude, lon=Longitude, mask=inps.mask, outDir=inps.workdir, outputFormat=inps.outputFormat, stitchMethodType='overlap', verbose=inps.verbose, num_threads=inps.num_threads, multilooking=inps.multilooking)
+    else:
+        inps.layers = []
```
ehavazli, line 311: isn't this else obsolete?

sssangha: @ehavazli no, it's needed for the last check, which generates an amplitude stack only if the user specifies the download of amplitude layers (either explicitly, or by specifying 'all' as an argument). Makes sense?

ehavazli (approved):
looks good to me

dbekaert:
Thanks Sim!

yunjunz:
#244

PR #247, opened 2021-01-03
add space to shebang

bbuzzanga:
After running setup.py on a fresh conda environment, the wrong environment gets specified in the Miniconda/bin/aria* files ... except for ariaDownload.py.  The only difference was a space after the #! in the shebang, and this has fixed the other ones. Very strange. 

#!/Users/bb/Miniconda3/envs/SL/bin/python -> incorrect without initial space in the source dir
#!/Users/bb/Miniconda3/envs/ARIA/bin/python -> correction after putting space.


bbuzzanga:
included instead in #249

Issue #248, opened 2021-01-07
[BUG] DEM not recognized


rzinke:
**Describe the bug**
ARIA-tools does not properly download and format the SRTM DEM files necessary.

**To Reproduce**
1. Download 
```ariaDownload.py -t 41 -s 20190114 -e 20190228 -w ARIAproducts```

2. Command used
```ariaTSsetup.py -f "ARIAproducts/*.nc" -w StitchedProducts```

3. Error Output
```
***Time-series Preparation Function:***
Multi-core version
All (70) GUNW products meet spatial bbox criteria.
Group GUNW products into spatiotemporally continuous interferograms.
All (5) interferograms are spatially continuous.
Thread count specified for gdal multiprocessing = 2
Download/cropping DEM
WARNING: User-defined bounds 611659km² supersedes DEM maximum download area of 225000km², must download in chunks
Traceback (most recent call last):
  File "/Users/rzinke/opt/miniconda3/envs/ARIA-tools/bin/ariaTSsetup.py", line 15, in <module>
    main(inps)
  File "/Users/rzinke/opt/miniconda3/envs/ARIA-tools/lib/python3.8/site-packages/ARIAtools/tsSetup.py", line 275, in main
    inps.demfile, demfile, Latitude, Longitude = prep_dem(inps.demfile, standardproduct_info.bbox_file, prods_TOTbbox, \
  File "/Users/rzinke/opt/miniconda3/envs/ARIA-tools/lib/python3.8/site-packages/ARIAtools/extractProduct.py", line 299, in prep_dem
    demfilename = dl_dem(aria_dem, prods_TOTbbox_metadatalyr, num_threads)
  File "/Users/rzinke/opt/miniconda3/envs/ARIA-tools/lib/python3.8/site-packages/ARIAtools/extractProduct.py", line 376, in dl_dem
    gdal.Warp(dst, chunked_files, options=gdal.WarpOptions(multithread=True, options=['NUM_THREADS=%s'%(num_threads)]))
  File "/Users/rzinke/opt/miniconda3/envs/ARIA-tools/lib/python3.8/site-packages/osgeo/gdal.py", line 618, in Warp
    srcDSTab.append(Open(elt))
  File "/Users/rzinke/opt/miniconda3/envs/ARIA-tools/lib/python3.8/site-packages/osgeo/gdal.py", line 3308, in Open
    return _gdal.Open(*args)
[...]

PR #249, opened 2021-01-08
correct DEM chunking download for large areas

bbuzzanga:
This addresses #248. 
I've fixed a number of problems, including the incorrect order of the SNWE coordinates and the calculation of the chunk size.
There may be one or two more related to how the DEM gets stitched back together after chunking though. @rzinke do you have any idea what the DEM should look like? i.e. probably not like the figure below.

image
Units on colorbar are meters, black is no data.


rzinke:
Thanks for taking a look into this, @bbuzz31. The DEM should look more like this: 

Screen Shot 2021-01-08 at 10 09 14 AM

bbuzzanga:
Thanks for the image. I've found the stitching problem and now produce the figure below, which I believe is correct. Unfortunately, I think there is another problem elsewhere as the time series results do not look correct. When you get there please open a new issue, perhaps with some details on where you think the bug occurs. 

In general, you may want to split up the track into a few pieces, process them separately, and them put them back together for time series processing. Using such a large area has not been rigorously tested in ARIA-tools (as you've discoverd) and you may run into, among others, memory problems on your machine at some point.

DEM_stitch

dbekaert:
@rzinke Can you verify this fixes the problem?

dbekaert:
@sssangha could you have a peak at this PR?

PR #252, opened 2021-02-11
Stitch new

dbekaert:


dbekaert:
@ehavazli Could you look into this code?

dbekaert:
retiring this as we have approach from marin now

PR #255, opened 2021-04-19
requirements: separate dependencies from ARIA-tools-docs

yunjunz:
This PR updates the `requirements.txt` file by removing the notebook dependencies for ARIA-tools-docs, so that one could use `ARIA-tools/requirements.txt` and `MintPy/docs/conda.txt` to install both packages in one environment, as shown here (https://github.com/yunjunz/conda_envs), which would be simplified by this PR as well. A short note on the dependency installation is added in the README.md file, as an alternative to the `conda env create` approach.

Since requirements.txt is not used anyway currently before this PR, this should not affect any funcationality.

The detailed changes are as below:

+ requirements: remove dependencies of notebooks for ARIA-tools-docs

+ env*.yml: separate dependencies for ARIA-tools-docs with comments

+ README: add notes for `conda install --file requirements.txt`, as an alternative to `conda env create`.

+ README: update links to the source build md files.

+ README: indentation adjustment in the contents

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,14 +21,14 @@ THIS IS RESEARCH CODE PROVIDED TO YOU "AS IS" WITH NO WARRANTIES OF CORRECTNESS.
 ## Contents
 1.  [Software Dependencies](#software-dependencies)
 2.  [Installation](#installation)
--   [Conda](#conda)
--   [Other installation options](#other-installation-options)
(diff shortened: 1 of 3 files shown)


dbekaert (approved):
LGTM @yunjunz  thanks!

PR #256, opened 2021-04-27
PEP8 styling for time series prep module

ehavazli:
This PR is my initial stab at PEP8 styling of the time series module. The changes are introduced in tsStep.py (new name ts_setup.py) and ariaTSsetup.py (wrt changes in module naming etc.)


bbuzzanga:
The new script looks good, but it's hard to follow the changes. What are all the merge branches thing? Can they be squashed?

Main concern is that naming changes need to be propagated to the docs repo.

ehavazli:
> The new script looks good, but it's hard to follow the changes. What are all the merge branches thing? Can they be squashed?
> 
> Main concern is that naming changes need to be propagated to the docs repo.

There are no changes to the code except moving things around but I can try to implement the name change as the last fix after every one is ok with the changes. I agree, the cascading effects on the docs side would require attention as well.

Other than that, working on the Merge branch commits, which shouldn't have been there to begin with, but rebase seem to misbehave. May delete this PR and issue a new one if can't resolve on this one.

PR #257, opened 2021-04-28
Apply PEP8 style changes

ehavazli:
trying this again

PR #254, opened 2021-03-16
Remove vestigial dependency

sssangha:

PR #261, opened 2021-05-03
Dependencies pep8

bbuzzanga:
this incorporates #254 and extends it by adding a required pyproj dependency to the ARIA-tools install file
it then incorporates  PEP8 style changes in #257.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ jobs:
             python setup.py build
             python setup.py install
             ariaDownload.py -h
-            prep_aria.py -h
+            # prep_aria.py -h
       - run:
           name: Run unit tests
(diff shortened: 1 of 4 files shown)


ehavazli:
Codacy issue is related with two rules in PEP257 (D212 and D213). These rules are exclusive and can be ignored as long as one of them is satisfied.

dbekaert (approved):

dbekaert:
@ehavazli avoid doing PEB8 changes on the unwrap stitching code. Better to do that on the version that has the 2 stager from gang on it. If you could test that code and get it into dev would be great.

dbekaert:
@bbuzz31 Looks good to me on the requirements.

PR #272, opened 2021-06-17
only download latest version of duplicate files

bbuzzanga:
For example, both of these products are currently available:
```
S1-GUNW-A-R-004-tops-20200307_20200224-230838-45104N_43235N-PP-b27e-v2_0_2
S1-GUNW-A-R-004-tops-20200307_20200224-230838-45104N_43235N-PP-b27e-v2_0_4
```
This PR will just count, and download the second (both physically and in the url file)

### DIFF
--- a/tools/ARIAtools/ARIAProduct.py
+++ b/tools/ARIAtools/ARIAProduct.py
@@ -21,7 +21,7 @@ gdal.PushErrorHandler('CPLQuietErrorHandler')
 
 log = logging.getLogger(__name__)
 
-# Unpacking class fuction of readproduct to be become global fucntion that can be called in parallel
+# Unpacking class fuction of readproduct to become global fucntion that can be called in parallel
 def unwrap_self_readproduct(arg):
     # arg is the self argument and the filename is of the file to be read
     return ARIA_standardproduct.__readproduct__(arg[0], arg[1])[0]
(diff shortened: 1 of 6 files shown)


dbekaert:
Excellent @bbuzz31. That would reduce some overhead in the ARIA-tools extraction step.

Do you perhaps want to make it a flag; e.g. fetch all; fetch latest; fetch a version?
This would allow for potential reproducing results of a paper etc.

sssangha:
@bbuzz31 looks good to me.
I just pushed one simple update to delete duplicate products that had already been downloaded and aren't used. If the products aren't deleted locally, then they may be passed along the ariaExtract/ariaTSsetup pipeline.

Let me know what you think, but with that it looks good to go.

bbuzzanga:
> @bbuzz31 looks good to me.
> I just pushed one simple update to delete duplicate products that had already been downloaded and aren't used. If the products aren't deleted locally, then they may be passed along the ariaExtract/ariaTSsetup pipeline.
> 
> Let me know what you think, but with that it looks good to go.

Thanks for having a look @sssangha. I don't think we should delete a user's data. I've changed it to move the extra duplicated products to a different location.

sssangha:
> Thanks for having a look @sssangha. I don't think we should delete a user's data. I've changed it to move the extra duplicated products to a different location.

@bbuzz31 True, we shouldn't be deleting the duplicates altogether. The user may not intend this so it indeed would be best to move it to a separate area.

So with that, everything lgtm!

dbekaert (approved):

Issue #273, opened 2021-06-23
[BUG] pkg_resources.DistributionNotFound for print version


yunjunz:
**Describe the bug**
Hi ARIA-tools team, it seems like the recent print version (#269) is not working for me. Attached below is the printout msg from ariaDownload.py, and several other (did not try all) scripts are having the same error as well.

FYI, I use the requirements.txt file to install the dependencies and manually setup the PYTHONPATH and PATH environment variable, without using setup.py.

If I commented out the following lines, everything is working fine without issue:
```
from pkg_resources import get_distribution
print ('ARIA-tools Version:', get_distribution('ARIAtools').version)
```

**To Reproduce**
1. Command used
```
ariaDownload.py -h
```
2. Error Output
```
(insar) yunjunz:~>$ ariaDownload.py -h
Traceback (most recent call last):
  File "/Users/yunjunz/tools/ARIA-tools/tools/bin/ariaDownload.py", line 473, in <module>
    print ('ARIA-tools Version:', get_distribution('ARIAtools').version)
  File "/Users/yunjunz/tools/miniconda3/envs/insar/lib/python3.7/site-packages/pkg_resources/__init__.py", line 480, in get_distribution
    dist = get_provider(dist)
  File "/Users/yunjunz/tools/miniconda3/envs/insar/lib/python3.7/site-packages/pkg_resources/__init__.py", line 356, in get_provider
    return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
  File "/Users/yunjunz/tools/miniconda3/envs/insar/lib/python3.7/site-packages/pkg_resources/__init__.py", line 899, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/Users/yunjunz/tools/miniconda3/envs/insar/lib/python3.7/site-packages/pkg_resources/__init__.py", line 785, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'ARIAtools' distribution was not found and is required by the application
```

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - ARIA-tools git tag: [e.g. git show --summary], the latest version from `upstream/dev` branch.
 - OS: macOS 10.14

**Additional context**

dbekaert:
@bbuzz31 could you look into this? 
I though you added a try except for it such in case of failure it would pass.

dbekaert:
@yunjunz how do you currently capture the Mintpy version?

[...]

PR #274, opened 2021-06-24
add try/except around get distribution

bbuzzanga:
#273


dbekaert (approved):
LGTM. At least it would not error out now!

Issue #265, opened 2021-05-20
[BUG] DEM download fails for large areas


bbuzzanga:
**Describe the bug**
ARIA area calculation for DEM extent to decide whether to chunk or not does not match the SRTM servers.
Relevant line is 349 in extractProduct.py

Note the area calculation (`shapefile_area`) is correct.

**To Reproduce**
1. `ariaDownload.py -t 004 -s 20191201 -e 20200101` # 13 products 
2. `ariaExtract.py  -f "./products/*.nc" -l all -w . -d Download`
```Traceback (most recent call last):
  File "/Users/bb/Miniconda3/envs/ARIA/bin/ariaExtract.py", line 16, in <module>
    main(inps)
  File "/Users/bb/Miniconda3/envs/ARIA/lib/python3.8/site-packages/ARIAtools/extractProduct.py", line 942, in main
    inps.demfile, demfile, Latitude, Longitude = prep_dem(inps.demfile,
  File "/Users/bb/Miniconda3/envs/ARIA/lib/python3.8/site-packages/ARIAtools/extractProduct.py", line 310, in prep_dem
    gdal.Warp(aria_dem, demfilename, format=outputFormat,
  File "/Users/bb/Miniconda3/envs/ARIA/lib/python3.8/site-packages/osgeo/gdal.py", line 665, in Warp
    srcDSTab = [Open(srcDSOrSrcDSTab)]
  File "/Users/bb/Miniconda3/envs/ARIA/lib/python3.8/site-packages/osgeo/gdal.py", line 4192, in Open
    return _gdal.Open(*args)
RuntimeError: `./DEM/SRTM_3arcsec_uncropped.tif' not recognized as a supported file format.
```

We calculate the area of `./productBoundingBox/productBoundingBox_croptounion_formetadatalyr.json` as 84497 km^2, which is correct, but `./DEM/SRTM_3arcsec_uncropped.tif`  reads:
`Error: The maximum area for SRTMGL1_E is 450,000 km2. The selected area is 472,979 km2`

bbuzzanga:
In #269 I put in a hack to initialize downloading in chunks if the area we calculate is greater than 80000 km^2, which solves the problem (for now at least).

rzinke:
Hi @bbuzz31, I am still getting an error from this after updating to the most recent version of AT and reinstalling.

```
$ git log
[...]

PR #275, opened 2021-07-19
Patch download and area calculation

sssangha:
A few bug-fixes were implemented:

1. In `ariaDownloads.py`, duplicate products (if they already existed) were originally not moved into separate subdirectory so they may not be passed along. This patch ensures such duplicate products are indeed segregated to this subdirectory.

2. Option added to `shapefile_area` function in `shapefile_util.py` to estimate the area of the region, as opposed to just the cutline, in order to obtain an accurate estimate of the extent of a downloaded DEM and/or mask. This resolves issue where DEMs weren’t chunked when they should have been, leading to crashes as the SRTM server cannot handle such large requests. Addresses #265 . 

3. Incremental updates to be inline with pep standards also passed.

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -27,9 +27,7 @@ log = logging.getLogger(__name__)
 _world_dem = "https://portal.opentopography.org/API/globaldem?demtype=SRTMGL1_E&west={}&south={}&east={}&north={}&outputFormat=GTiff"
 
 def createParser():
-    '''
-       Extract specified product layers. The default will export all layers.
-    '''
+    """Extract specified product layers. The default will export all layers."""
     import argparse
     parser = argparse.ArgumentParser(description= \
             'Program to extract data and meta-data layers from ARIA standard GUNW products.'\
@@ -76,14 +74,10 @@ def cmdLineParse(iargs = None):
     return parser.parse_args(args=iargs)
 
 class InterpCube(object):
-    '''
-        Class to interpolate intersection of cube with DEM
-    '''
+    """Class to interpolate intersection of cube with DEM."""
 
     def __init__(self, inobj, hgtobj, latobj, lonobj):
-        '''
-            Init with h5py dataset.
-        '''
+        """Init with h5py dataset."""
         self.data = inobj[:]
         self.hgts = hgtobj[:]
         self.offset = None
@@ -94,9 +88,7 @@ class InterpCube(object):
         self.createInterp()
 
     def createInterp(self):
-        '''
-            Create interpolators.
-        '''
+        """Create interpolators."""
         from scipy.interpolate import RectBivariateSpline
         self.offset = np.mean(self.data)
         for i in range(len(self.hgts)):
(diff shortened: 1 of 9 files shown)


tools/ARIAtools/extractProduct.py
```diff
@@ -387,20 +378,24 @@ def prep_dem(demfilename, bbox_file, prods_TOTbbox, prods_TOTbbox_metadatalyr,
...
     from ARIAtools.shapefile_util import shapefile_area
 
     root      = os.path.splitext(path_dem)[0]
     prod_shapefile = open_shapefile(path_prod_union, 0, 0)
     WSEN      = prod_shapefile.bounds
-    # If area > 225000 km2, must split requests into chunks to successfully access data
+    # If area > 450000 km2, must split requests into chunks to successfully access data
     chunk = False
-    if shapefile_area(prod_shapefile) > 300000:
+    if shapefile_area(prod_shapefile, bounds = True) > 450000:
```
dbekaert: @sssangha suggest to be conservative and make this sligthly smaller than the openTopo API.

dbekaert:
```suggestion
    if shapefile_area(prod_shapefile, bounds = True) > 400000:
```

tools/ARIAtools/extractProduct.py
```diff
@@ -387,20 +378,24 @@ def prep_dem(demfilename, bbox_file, prods_TOTbbox, prods_TOTbbox_metadatalyr,
...
         chunk = True
         # Increase chunking size to discretize box into smaller grids
-        log.warning('User-defined bounds results in an area of %d km which exceeds the maximum download area of 450000; downloading in chunks', shapefile_area(prod_shapefile))
+        log.warning('User-defined bounds results in an area of %d km which ' \
+                    'exceeds the maximum download area of 450000; ' \
```
dbekaert:
```suggestion
                    'exceeds the maximum download area of 400000; ' \
```

dbekaert:
@rzinke @bbuzz31 Can you test this over your region to see if it remove the issue on the DEM download?

jlmaurer:
@sssangha I'm still having the same issue even after a clean start and pulling #275. DEM has all zeros and files that are not VRT files have no valid pixels (besides the unwrapped phase). This smells like it could be a GDAL issue, I'm using version:
`gdal                      3.1.0            py38h9f7df5a_3    conda-forge`
Could this be an issue?

jlmaurer:
Note that amplitude and coherence are VRTs and they work fine, have valid pixels etc.

sssangha:
@dbekaert these updates now ensure that regardless of what executable the user launches (e.g. `ariaDownload.py`, `ariaExtract.py`, `ariaTSsetup.py`, and `ariaPlot.py`), duplicate products are captured and quarantined to a separate subfolder. I also cleaned up the code by moving this function out of `ariaDownload.py` and into a separate util script that is accessible to other scripts.

@bbuzz31 your contributions look good on my end! Was able to perform tests with a large DEM without a problem.

@jlmaurer I do recall issues with some relatively recent gdal update, but I was again unable to replicate your issue with `gdal 3.1.2`. Could you please retry again by updating to the specific version I vetted this code with? I.e.: `conda install gdal=3.1.2`.

If it is indeed an issue with older versions of gdal as I suspect, we may update the environment file accordingly.

bbuzzanga (approved):
GACOS update LGTM

PR #277, opened 2021-07-29
Update README.md

bbuzzanga:
typo fix


bbuzzanga:
circleci is failing to download the watermask because the server at UH hosting the GSHHG shapefiles is down

dbekaert (approved):

PR #280, opened 2021-08-02
Relative path patch

sssangha:
Following PR #160 and the advice of @yunjunz from #253, absolute path designations reverted back to relative path.

@ehavazli @bbuzz31 as a sanity check, do you recall any specific reason from previous discussions regarding why an absolute path would be preferred over relative paths?

### DIFF
--- a/tools/ARIAtools/tsSetup.py
+++ b/tools/ARIAtools/tsSetup.py
@@ -335,7 +335,7 @@ def generate_stack(aria_prod, input_files, output_file_name,
                 print('Orbit direction not recognized')
                 metadata['orbit_direction'] = 'UNKNOWN'
 
-            path = os.path.abspath(data[1])
+            path = os.path.relpath(os.path.abspath(data[1]), start=stack_dir)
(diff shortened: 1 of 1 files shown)


dbekaert:
Thanks @yunjunz, does this PR address your issues?

@sssangha: If green lighted by others can you:
1) backport the change to the release-v1.1.2 branch
2) make a re-release of the v1.1.2 release from the v1.1.2 branch.

bbuzzanga (approved):

ehavazli:
> Following PR #160 and the advice of @yunjunz from #253, absolute path designations reverted back to relative path.
> 
> @ehavazli @bbuzz31 as a sanity check, do you recall any specific reason from previous discussions regarding why an absolute path would be preferred over relative paths?

I believe this change was applied to make MintPy processing more portable/flexible. If it is defeating the purpose, there is no reason to stick with abspath.

ehavazli (approved):

yunjunz:
> Thanks @yunjunz, does this PR address your issues?

Yes, thank you @dbekaert and @sssangha.

yunjunz (approved):

PR #281, opened 2021-08-02
backport to the released version 1.1.2

dbekaert:


dbekaert:
backport and re-release has been done.

@yunjunz @ehavazli should be good to go now.
I think it might require a shutdown of instance on OSL and re-start to pull new changes.

PR #284, opened 2021-10-19
Some tweaks to allow editable installs

jhkennedy:
This changes `setup.py` to allow for develop/editable installs, which create a symbolic package pointing back to the repository and allows code changes to be automatically reflected in the environment. This is particularly useful for debugging. You can install in editable mode with the command:
```
# preferred
python -m pip install -e .
# depreciated
python setup.py develop
```
from the repository root. 

I've also:
* updated the install instructions in the README to prefer pip over directly calling `setup.py` which is preferred as pip handles dependency resolution and installation nuances better than vanilla setuptools. 
* Added a more comprehensive `.gitignore`, generated by [Joe](https://github.com/karan/joe) (no relation), that covers Python, Jupyter, Jetbrains IDEs (e.g. PyCharm) and vim. I can back this out if preferred.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,255 @@
-*.pyc
-*~
-*.swp
-*.DS_Store
-__pycache__
-.ipynb_checkpoints
+#### joe made this: http://goel.io/joe
+
+#### python ####
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.nox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*.cover
+*.py,cover
+.hypothesis/
+.pytest_cache/
+cover/
+
+# Translations
+*.mo
+*.pot
+
+# Django stuff:
 *.log
-build
+local_settings.py
+db.sqlite3
+db.sqlite3-journal
+
+# Flask stuff:
+instance/
+.webassets-cache
+
+# Scrapy stuff:
+.scrapy
+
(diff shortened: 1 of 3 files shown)


sssangha:
Thanks @jhkennedy!

I'm trying to test this PR myself though and I'm running into an issue. Can you perhaps advise?
```
(ARIA-tools) [ssangha@sar ARIA-tools]$ python -m pip install .
Processing /u/sar-r2/ssangha/conda_installation/stable_aug14_2020/ARIA-tools
  Preparing metadata (setup.py) ... done
Building wheels for collected packages: ARIAtools
  Building wheel for ARIAtools (setup.py) ... done
  Created wheel for ARIAtools: filename=ARIAtools-1.1.0-py3-none-any.whl size=123025 sha256=103ced0551ac3b34bf5102ffcd8f51f469c919deb9371874d32ffc46b5c12d0a
  Stored in directory: /u/sarh0/ssangha/.cache/pip/wheels/b3/b1/df/9e154ac32375415f052776ccb40055dcedf35851ae515fcb56
Successfully built ARIAtools
Installing collected packages: ARIAtools
  Attempting uninstall: ARIAtools
    Found existing installation: ARIAtools 1.0.0
ERROR: Cannot uninstall 'ARIAtools'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
(ARIA-tools) [ssangha@sar ARIA-tools]$ pwd
/u/sar-r2/ssangha/conda_installation/stable_aug14_2020/ARIA-tools
```

Is this something to be expected when there is an existing ARIA-tools install? If so, we should figure out a way to handle such cases as it would be a struggle to have to uninstall and manage the environment with each PR pull?

Thanks again

jhkennedy:
@sssangha you're correct, if you have an environment that's been created with
```
python setup.py install # or develop
```
pip won't be able to uninstall/modify the environment correctly because setuptools/distutils doesn't write all the necessary metadata pip needs. With a setuptools/distutils install you _have_ to remove installed files manually (another reason to prefer pip). 

(setuptools/distutils doesn't care if stuff is already installed, it doesn't check and just overwrites, so you can end up with stale files installed as well if there were renames/moves/etc.)

The best bet in that case it to remove the conda environment, and re-install using pip (you can clone the old one first to preserve it if you want). Or for testing, use a different environment.

jhkennedy:
@sssangha once the environment is set up and things are installed with `pip`, you can easily reinstall/uninstall etc with pip. 

Typical pattern for keeping the environment and install up to date for a developer would look like:
```
conda env update -f environment.yml
python -m pip install -e .
```

the conda command is only necessary is dependencies (environment) has changed, and the pip command is only necessary if the package structure has significantly changed (new entrypoints, scripts, moved files, etc). But, there's no harm in running those two commands at anypoint to ensure everything is up to date.

bbuzzanga:
This is great! Thanks @jhkennedy

bbuzzanga (approved):

dbekaert (approved):
Thanks @jhkennedy

PR #286, opened 2021-11-10
bug fix on subsetting

bbuzzanga:
There is ambiguity in the start and stop time in the scene information fetched with ASF search, as the times only refer to the reference acquisition in the pair.

I didn't catch this in the original push, so the subsetting options didn't work correctly.

Also had the choice to always do one loop over the products, or sometimes do multiple loops. Opted for the former.

### DIFF
--- a/tools/bin/ariaDownload.py
+++ b/tools/bin/ariaDownload.py
@@ -52,10 +52,10 @@ def createParser():
                         type=str,
         help='Specify directory to deposit all outputs. Default is '
              '"products" in local directory where script is launched.')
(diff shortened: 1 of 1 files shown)


dbekaert:
@bbuzz31 thanks for catching this. 

We should also open an issue ticket on the ASF DAAC API page. @jhkennedy might be able to provide some guidance how to best proceed on the ASF DAAC side. Perhaps this is either something they can handle on the API side or might be something that could be fixed on the CMR side. @jhkennedy migth know the right people to reach out to.

dbekaert (approved):

PR #290, opened 2022-01-07
Bb gdalconfig

bbuzzanga:
implements a new gdalconfig option to (hopefully) speed up virtual data processing


sssangha:
Ok, so I pulled out the repeated re-opening and redundant access of DEM metadata from the `finalize_metadata` function and instead pass a DEM after extracting everything we need from the top level. I hope this makes a tangible difference.

Plus did some house cleaning in the relevant functions to bring it up to PEP-8 standards and hopefully more readable

dbekaert:
@sssangha @bbuzz31 Lets continue with this work and go towards refactoring. Please capture processing time changes ad well as a statement comparing past and new outputs to capture and avoid potential errors introduced.

Issue #292, opened 2022-01-24
[BUG] DEM download now requires an API Key


bbuzzanga:
The changes upstream at open Topo prevent us from downloading SRTM dems without an API Key.
- See: https://opentopography.org/developers

We now have to add the custom API_Key to the URL string of our DEM, i.e.,:
```
_world_dem = "https://portal.opentopography.org/API/globaldem?demtype="\
               "SRTMGL1_E&west={}&south={}&east={}&north={}&outputFormat=GTiff"
_world_dem += '&API_Key=yourapikey'
```
This is no problem.

However, we will need to decide how we want the user to store their keys for us to access them.
It seems the common secure ways are to 
1) store the key in a file 
2) store it in an environment variable.

Either way is fine I think. Copernicus data service uses the former. 
Does anyone have any strong opinions? We will need to update the installation documents pronto.

sssangha:
PR #293  implements the former option, specifically aria-tools now accesses the API key from a `~/.topoapi` file... The installation section of the readme is also updated to reflect definition of this `~/.topoapi` file

PR #293, opened 2022-01-26
Support OpenTopo API change

sssangha:
Update to reflect OpenTopo API access update and successfully access DEMs. Addresses #292 

Will confirm here once I update installation instructions on main page to reflect necessary, prerequisite definition of hidden topoapi file.


sssangha:
Confirmed! Please refer to the installation section of the readme

bbuzzanga:
@sssangha LGTM.

Any idea what we should do about circleci?

FYI; if you don't want to import a new library (pathlib) just to get the home directory, you can do `os.expanduser('~')`

bbuzzanga (approved):
I think we're good to go now.

PR #296, opened 2022-03-23
Stitching warp2transform

rzinke:
Switched out gdalWarp functions used to apply a cutline to the unwrapped files, with gdalTranslate functions to crop the unwrapped files. Applied a mask band generated for the connComp files to mask out non-overlapping areas.
Some Pep8 style updates.

PR #298, opened 2022-04-06
Misclosure verbosity fix

rzinke:
Fixed issues with verbosity attribute. While fixing that, I discovered an issue with accepting and parsing the query point argument. I fixed that, too, and updated the code styling.


bbuzzanga:
@rzinke I think the bug fixes are incorporated elsewhere (e.g. #306) and the rest of this is incorporated in #296? Please reopen if not.

PR #299, opened 2022-04-16
Profiling

bbuzzanga:
Jupyter notebook for testing physical / virtual processing.

Untested, as aria-tools install on OSL is non-trivial.


bbuzzanga:
Will deal with this locally until finalized

PR #314, opened 2022-07-28
dont access stats

bbuzzanga:
this should allow the newest gdal versions to work, and save time in virtual processing.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -11,7 +11,7 @@ dependencies:
   - python>=3.6
   - asf_search
   - cartopy
-  - gdal>=3.5
+  - gdal>=3.5.1
   - hdf5
   - joblib
   - libgdal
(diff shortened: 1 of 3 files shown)


sssangha:
Updated to pass bperp mean value, as expected by prep_aria.py (https://github.com/insarlab/MintPy/blob/main/mintpy/prep_aria.py#L584), but with the means of the other geometry files discarded as they are no longer needed.

In my tests, we are able to circumvent the gdalstats issue to access this info efficiently by forcing an install of gdal >=3.5 (see gdal ticket here: https://github.com/OSGeo/gdal/issues/6023)

sssangha:
@bbuzz31 it was brought to my attention earlier this week that there is an issue trying to access NLCD due to server issues. Hence, the circleci failure is to be expected

sssangha (approved):
LGTM

PR #316, opened 2022-08-09
sync requirements.txt against environment.yml

yunjunz:
+ environment.yml:
   - install h5py via conda, instead of pip
   - remove hdf5 as it's not directly used in the code, and will be installed through h5py
   - sort by alphabetic order, for easy navigation

+ requirements.txt: sync against the environment.yml file

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -18,6 +18,7 @@ dependencies:
   - matplotlib
   - netcdf4
   - pandas
+  - pip
   - pyproj
   - requests
   - scipy
--- a/environment.yml
+++ b/environment.yml
@@ -11,20 +11,17 @@ dependencies:
   - python>=3.6
   - asf_search
   - cartopy
-  - libgdal>=3.5.1
   - gdal>=3.5.1
-  - hdf5
+  - libgdal>=3.5.1
+  - h5py
   - joblib
   - matplotlib
   - netcdf4
   - pandas
-  - pip
-  - requests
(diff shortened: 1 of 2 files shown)


bbuzzanga:
we need `pip` to handle the install of ariaTools, it's installed by default though right?

yunjunz:
> we need `pip` to handle the install of ariaTools, it's installed by default though right?

Yes, it will be installed while running `conda install --file requirements.txt`. Since it's used in the installation note later on, I added it back anyway to be explicit.

yunjunz:
@bbuzz31 if `ARIA-tools` works in python 3.10, I think it's best to NOT add `python <3.10` in the requirement files (https://github.com/aria-tools/ARIA-tools/pull/315). Whoever is installing isce2 and ARIA-tools should pin `python<3.10` in their own installation note.

bbuzzanga:
> @bbuzz31 if `ARIA-tools` works in python 3.10, I think it's best to NOT add `python <3.10` in the requirement files (#315). Whoever is installing isce2 and ARIA-tools should pin `python<3.10` in their own installation note.

hmm perhaps. But this release is intended to ensure consistency with ISCE2. If someone is installing everything on their own in the wrong order they might be in for unwanted surprises. What do you think @yunjunz @sssangha @ehavazli

bbuzzanga (approved):

PR #318, opened 2022-08-10
tsSetup bperp bug-fix

sssangha:
Pass mean of bperp file, not unw files.


sssangha:
Thanks for finding the bug and testing the fix @bbuzz31

bbuzzanga (approved):
Thanks @sssangha!

PR #319, opened 2022-08-10
Correct dupes

bbuzzanga:
Previously, we identified new version products based on the first ten fields of the file name (see [Product Filename Convention](https://aria.jpl.nasa.gov/products/standard-displacement-products.html)). 

However, the newer version products have different Decimal latitude than previous versions, despite covering the same area. 

For now there are two options available to the user:

1.  (default): all products are downloaded regardless of version number. 
    - Duplicates are identified by opening the file and comparing the timing and and area. 
    - The newest version is used for processing.
    
2. A user can specify a product version, and ONLY products which match that version are used.


sssangha (approved):
LGTM! Thanks @bbuzz31

PR #320, opened 2022-08-11
UNAVCO 2022

bbuzzanga:
Consistent environments with MintPy, isce2, for UNAVCO 2022

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,11 +8,11 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - python>=3.6
+  - python>=3.6,<3.10
   - asf_search
   - cartopy
-  - gdal>=3.5.1
-  - libgdal>=3.5.1
+  - gdal<=3.4.2
+  - libgdal>=3.4.2
   - h5py
   - joblib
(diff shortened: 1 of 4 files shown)


sssangha (approved):
Everything looks great!

environment.yml
```diff
@@ -8,11 +8,11 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - python>=3.6
+  - python>=3.6,<3.10
   - asf_search
   - cartopy
-  - gdal>=3.5.1
-  - libgdal>=3.5.1
+  - gdal<=3.4.2
```
yunjunz, line 14: Should this be `gdal>=3.4.2`?

sssangha: We resolved this in the latest patch

bbuzzanga: @yunjunz actually no, because we want to have a consistent environment with ISCE, which at my last check (a couple days ago) was restricted to gdal=3.4.2

yunjunz: `gdal` version 3.4.2 is used by mamba from my installation as well.

PR #323, opened 2022-08-16
gdal<=3.4.2 to match ISCE

bbuzzanga:


sssangha (approved):
LGTM. @yunjunz , I believe this sorts out some incompatibility vis-a-vis ISCE

yunjunz:
I don't think this is the right change, IMHO. It does not break the incompatibility for now, but will quickly do in the near future. 

Similar to my previous comments on python 3.10: ARIA-tools may be better off with version constraints based on its own code. It was isce2 that should add the version constraint of python<3.10, not ARIA-tools.

Even without the change from this PR, conda / mamba already figured out that gdal 3.4.2 is the version to be used. So I don't think this PR is needed.

In general, I would be cautious about putting a min version constraint that is recent, and would be even more cautious of putting a max version constraint.

Following this spirit, the only constraint I would think ARIA-tools needs for gdal is a min version constraint that equals to the lowest possible version that ARIA-tools could work correctly.

PR #325, opened 2022-08-18
adjust regex to only search version string

bbuzzanga:
currently the whole string is searched for version starting with 'v' and fails if there is e.g., a username joyvan


sssangha (approved):
LGTM

PR #337, opened 2023-01-11
add tropoTotal

mgovorcin:
add function to create troposphereTotal on the fly using gdal PixelFunction

did little clean up, please feel free to reject suggestions

### DIFF
--- a/tools/ARIAtools/ARIAProduct.py
+++ b/tools/ARIAtools/ARIAProduct.py
@@ -80,7 +80,7 @@ class ARIA_standardproduct:
     import glob
 
     def __init__(self, filearg, bbox=None, workdir='./', num_threads=1,
-                 url_version='None', nc_version='1c', verbose=False):
+                 url_version='None', nc_version='None', verbose=False):
         """
 
         Parse products and input bounding box (if specified)
--- a/tools/ARIAtools/ARIAProduct.py
+++ b/tools/ARIAtools/ARIAProduct.py
@@ -1,11 +1,11 @@
 #!/usr/bin/env python3
-#~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #
 # Author: Simran Sangha & David Bekaert
 # Copyright 2019, by the California Institute of Technology. ALL RIGHTS
 # RESERVED. United States Government Sponsorship acknowledged.
 #
-#~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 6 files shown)


tools/ARIAtools/ARIAProduct.py
```diff
@@ -77,8 +78,9 @@ class ARIA_standardproduct:
     spatiotemporally contiguous interferograms.
     """
     import glob
+
     def __init__(self, filearg, bbox=None, workdir='./', num_threads=1,
-                 url_version='None', nc_version='None', verbose=False):
+                 url_version='None', nc_version='1c', verbose=False):
```
mgovorcin: nc_version was giving me error so I set it to be the latest initially

sssangha:
@mgovorcin this PR looks solid, thanks for the integration of the discussed features and clean-up! I was able to get it working on my end with some minor fixes.

I have a few of broader notes:
I removed the function where you computed the total tropo within `tsSetup,py`, as this is handled in `extractProduct.py`, and without forcing the user to extract the dry and wet components when they don't want to.

To extract the total tropo by default though (i.e. without the user having to specify it as an argument), I reintroduced the `layerCheck` function and made some adjustments to the logic. This is intended to check whether or not the tropo layers exist in all standard products in the first place before proceeding.

Hence because of this, I made some changes to the way the default layers are managed and extracted.

tools/ARIAtools/tsSetup.py
```diff
@@ -205,52 +238,46 @@ def generate_stack(aria_prod, input_files, output_file_name,
...
 
-    if input_files in ['unwrappedPhase', 'unwrapped', 'unw']:
+    # MG; did a little clean up below
+
+    # MG: find pattern in string, with special case for tropo
+    # take into account variation in input name
+    # e.g. [unwrappedPhase, unwrapPhase, unwrap]
+    # NOTE is it necessary to have the option for different names
+    # as they are hardcoreded, maybe because of different nc version, double-check?
+    _find_match = lambda x: re.findall('[A-Z][^A-Z]*', x)[0] \
```
mgovorcin, line 251: This  could be moved to a separate script that check the nc version, input files, find stored layers in the container and  output the layers names for futher processing. Maybe for a next update :D

tools/ARIAtools/tsSetup.py
```diff
@@ -205,52 +238,46 @@ def generate_stack(aria_prod, input_files, output_file_name,
...
+                            if re.search('tropo', x) \
+                            else re.findall('[a-z][^A-Z]*', x)[0]
+
+    # Special case - gacos correction files
```
mgovorcin, line 255: Do not know Sim is it better to move older version of code for correcting unwrapped Phase with total tropo delay from GACOS to separate script:
1.  ingest GACOS products, and prepare tropoTotal vrt files (in a similar manner like ariaExtract.py)
2. use tsSetup.py to prepare tropoTotal stack vrt
3. have optional script to treat unwrappedPhase files with GACOS/other model tropoTotal files

mgovorcin: This is simplify both ariaExtract and tsSetup scripts,

mgovorcin (commented):
@sssangha  looks good! Added two comments for a next update

tools/ARIAtools/tsSetup.py
sssangha: @mgovorcin I understood that the intention was though to no longer directly correct the phase through ARIA-tools. I.e. once I merge these changes, we should work to now push corrections to a separate cube which we then interface with MintPy to use instead to correct the phase on their end.

sssangha (approved):

PR #338, opened 2023-01-12
BB interp

bbuzzanga:
ARIA tools was previously not intersecting with the DEM due to a small bug that overwrot the DEM values with longitude values instead.
That has been fixed and improved using an xarray-based implementation. The topography is now evident in the height layers (compare with https://github.com/dbekaert/RAiDER/pull/465):
image

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -17,6 +17,9 @@ dependencies:
   - matplotlib
   - netcdf4
   - pandas
+  - xarray
+  - rasterio
+  - rioxarray
   - pyproj
   - requests
   - scipy
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -135,7 +135,6 @@ class InterpCube(object):
 
 class metadata_qualitycheck:
     """Metadata quality control function.
-
     Artifacts recognized based off of covariance of cross-profiles.
     Bug-fix varies based off of layer of interest.
     Verbose mode generates a series of quality control plots with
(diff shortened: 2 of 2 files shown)


sssangha:
@dbekaert @bbuzz31 , below I’ll share the results of a basic test I ran to vet this bug-fix. Specifically, I compared the incidence angles derived directly through ISCE2 with those derived from the ARIA-tools interpolation and DEM-intersection scheme.

For reference, the example I used an ascending track (path 72) GUNW over Iraq: `S1-GUNW-A-R-072-tops-20171117_20171111-145939-00043E_00034N-PP-428e-v2_0_6.nc`

Here is the hillshaded view of the DEM:
Screen Shot 2023-01-17 at 1 36 21 PM

Here is the incidence angle raster derived directly from ISCE2:
Screen Shot 2023-01-17 at 1 39 04 PM


Note that it’s very difficult to discern the difference looking at this side-by-side with the ARIA-tools derived counterpart, so here’s instead the residual (i.e. ARIA-tools derived interpolated incidence angle minus the ISCE2 derived incidence angle). The residual looks like a DEM in and of itself, which would make sense if the topography doesn’t cancel out as we’d expect:
Screen Shot 2023-01-17 at 1 37 53 PM

After these proposed bug-fixes, here is the new residual between the “corrected” ARIA-tools and ISCE2 derived rasters. Clearly looks like the topography is cancelled out. The remaining topography-correlated residual is to be expected given the ARIA-tools field has been interpolated:
Screen Shot 2023-01-17 at 1 38 11 PM

PR #339, opened 2023-01-19
comment out lines for tropo to work

bbuzzanga:
rewrite the full tropo extraction

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -21,6 +21,7 @@ from ARIAtools.unwrapStitching import product_stitch_overlap, product_stitch_2st
 from ARIAtools.vrtmanager import resampleRaster, layerCheck
 import pyproj
 from pyproj import CRS, Transformer
+import xarray as xr
 import rioxarray as xrr
 import rasterio as rio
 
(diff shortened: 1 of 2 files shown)


tools/ARIAtools/extractProduct.py
```diff
@@ -664,59 +666,57 @@ def export_products(full_product_dict, bbox_file, prods_TOTbbox, layers,
...
-            #If necessary, resample raster
-            if multilooking is not None:
-                resampleRaster(outname, multilooking, bounds, prods_TOTbbox,
-                               rankedResampling, outputFormat=outputFormat,
-                               num_threads=num_threads)
+                t_comp     = [path_nc_wet.replace(tropo_prefix+start_key, tropo_prefix+key)]
+
+                hgt_field = prep_metadatalayers(t_outname, t_comp, dem)
+
+                finalize_metadata(t_outname, bounds, dem_bounds, \
```
sssangha: @bbuzz31 this procedure is now interpolating the intermediate tropo layers even if the user may not want them in the end. This introduces a bit of extra overhead we could've circumvented as I'd intended to do before, where I initialized a total tropo array using temp files before then interpolating just that directly.

sssangha:
Following a recent [PR](https://github.com/aria-tools/ARIA-tools/pull/338), I have confirmed that the behavior of interpolation/intersection is stable. Specifically, the residual between ISCE-derived full resolution incidence angles and ARIA-derived interpolated incidence angles is small and to be expected interpolation for latter:
Screen Shot 2023-01-20 at 3 31 59 PM

This is as opposed to residual before the bug-fix in [PR](https://github.com/aria-tools/ARIA-tools/pull/338) where the topography didn't cancel out as expected given the ARIA program was ignoring topography when intersecting:
image

PR #340, opened 2023-01-23
update GACOS function to agree with MintPy's

bbuzzanga:
This changes the resampling algorithm to use the default for interpolating the GACOS product to the ARIA crop. It also removes some unnecessary masking. After these changes, we agree within numerical precision to the GACOS correction written with MintPy (note I had to do some internal hacking to output a single SLC date and inverted units to compare with MintPy from aria tools)

After this PR: 
Screenshot 2023-01-22 at 16 36 54

Before this PR:
Screenshot 2023-01-21 at 15 01 15

Units are in meters.

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -1219,7 +1219,7 @@ def gacos_correction(full_product_dict, gacos_products, bbox_file,
             scale         = float(metadata_dict[1][i][0]) / (4*np.pi)
             tropo_product /= scale
 
-            # Account for lookAngle
+            # Account for incAngle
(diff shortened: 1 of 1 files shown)


dbekaert:
Is this using the incidence angle in aria tools rather than look angle?

bbuzzanga:
Yes this uses the incidence angle

dbekaert (approved):
LGTM

sssangha (approved):
All looks good on my end

dbekaert:
@sssangha the unit test seems to fail. This might be more related to something else that you are fixing on your end.

PR #341, opened 2023-01-25
fix incorrect difference and conversion in GACOS correction

bbuzzanga:
before fix, reference is subtracted from secondary and the conversion from meters to radians is off by a factor of -1


dbekaert (commented):
Does that not cancel out?

factor * (A-B) = -1 * factor*(B-A)

bbuzzanga:
Hmm yes, weird.

bbuzzanga:
Before:
image

After: 
image

PR #342, opened 2023-02-14
Fix environment and parsing bug

sssangha:
Patch env file to reflect scipy dependency restriction.

Also fix bug that did not properly handle cases where the user wishes to just extract the bounding boxes.


dbekaert:
@bbuzz31 @jlmaurer  Can you check in the unit test failure prior to merging

PR #348, opened 2023-03-04
Add support for sec and ref scenes

sssangha:
Add comprehensive check to parse and manage secondary and reference layers from embedded tropo and SET layers.

### DIFF
--- a/tools/ARIAtools/ARIAProduct.py
+++ b/tools/ARIAtools/ARIAProduct.py
@@ -17,6 +17,7 @@ from osgeo import gdal
 from ARIAtools.url_manager import url_versions
 from ARIAtools.shapefile_util import open_shapefile, save_shapefile
 from ARIAtools.logger import logger
+from ARIAtools.ARIA_global_variables import ARIA_TROPO_INTERNAL
 
 gdal.UseExceptions()
 gdal.PushErrorHandler('CPLQuietErrorHandler')
@@ -433,22 +434,25 @@ class ARIA_standardproduct:
                 '/science/grids/imagingGeometry/azimuthAngle'
             ]
             if version.lower()=='1c':
-                # get weather model name
-                meta = gdal.Info(fname)
-                for i in meta.split():
-                    if '/science/grids/corrections/external/troposphere/' in i:
-                        model_name = i.split('/')[-3]
-                        break
                 lyr_pref = '/science/grids/corrections'
                 sdskeys_addlyrs = [
                     lyr_pref + '/derived/ionosphere/ionosphere',
                     lyr_pref + '/external/tides/solidEarth'
-                        '/reference/solidEarthTide',
-                    lyr_pref + f'/external/troposphere/{model_name}'
-                        '/reference/troposphereWet',
-                    lyr_pref + f'/external/troposphere/{model_name}'
-                        '/reference/troposphereHydrostatic'
+                        '/reference/solidEarthTide'
                 ]
+                # get weather model name(s)
+                meta = gdal.Info(fname)
+                model_name = []
+                for i in meta.split():
+                    if '/science/grids/corrections/external/troposphere/' in i:
+                        model_name.append(i.split('/')[-3])
+                model_name = list(set(model_name))
+                for i in model_name:
+                    sdskeys_addlyrs.append(lyr_pref +
+                        f'/external/troposphere/{i}/reference/troposphereWet')
+                    sdskeys_addlyrs.append(lyr_pref +
+                        f'/external/troposphere/{i}/reference/'
+                        'troposphereHydrostatic')
                 # remove keys not found in product
                 sdskeys_addlyrs = [i for i in sdskeys_addlyrs if i in meta]
                 sdskeys.extend(sdskeys_addlyrs)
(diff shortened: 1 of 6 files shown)


sssangha:
For SET and tropo layers which contain corresponding reference and secondary scenes, the program organizes individual epochs within a `dates` subdirectory below the parent directory where differential fields are computed and deposited, as so:
```
(ARIA-tools) [ssangha@leffe just_SET2]$ ls
DEM  productBoundingBox  solidEarthTide  user_bbox.json
(ARIA-tools) [ssangha@leffe just_SET2]$ ls solidEarthTide/
20211120_20211108  20211120_20211108.aux.xml  20211120_20211108.vrt  20211120_20211108.xml  dates
(ARIA-tools) [ssangha@leffe just_SET2]$ ls solidEarthTide/dates/
20211108  20211108.aux.xml  20211108.vrt  20211108.xml  20211120  20211120.aux.xml  20211120.vrt  20211120.xml
```

sssangha:
@mgovorcin and @bbuzz31 next week I could really use your help to vet and expand upon this in `ariaTSsetup` as we work to interface with and update `prep_aria`. Specifically we need `ariaTSsetup` to setup a stack VRTs pointing to the date subdirectories as opposed to differentials, where the differentials eventually will be only generated if a user runs `ariaExtract` because we don't need/care about those for TS mode.

There are currently two sets of sample products for which the SET and tropo reference/secondary scenes have been inserted, respectively.

For SET refer to the products here: `/u/leffe-data/ssangha/solid_earth_test/A064/products`
And for testing purposes here is a minimum commanline input:
`ariaExtract.py -l 'solidEarthTide' -nt 20 -f "products/*.nc" -d Download -of ISCE --nc_version 1c -w just_SET`

For tropo refer to the products here: `/u/leffe-data/ssangha/solid_earth_test/temp_tropo_test/D071/products`
And for testing purposes here is a minimum commanline input:
`ariaExtract.py -l 'troposphereTotal' -nt 20 -f "products/*.nc" -d Download -of ISCE --nc_version 1c -w just_tropo`

Note @bbuzz31 that I've encountered the following error in this test stack you shared that I didn't see before. Could you please advise? Thanks:
```
>>> from osgeo import gdal
>>> gdal.BuildVRT('/u/leffe-data/ssangha/solid_earth_test/temp_tropo_test/D071/just_tropo/troposphereWet/HRES/dates/20200112', ['NETCDF:"/u/leffe-data/ssangha/solid_earth_test/temp_tropo_test/D071/products/S1-GUNW-D-R-071-tops-20200130_20200112-135156-34956N_32979N-PP-dd0c-v2_0_4.nc":/science/grids/corrections/external/troposphere/HRES/reference/troposphereWet'])
Warning 6: gdalbuildvrt does not support ungeoreferenced image. Skipping NETCDF:"/u/leffe-data/ssangha/solid_earth_test/temp_tropo_test/D071/products/S1-GUNW-D-R-071-tops-20200130_20200112-135156-34956N_32979N-PP-dd0c-v2_0_4.nc":/science/grids/corrections/external/troposphere/HRES/reference/troposphereWet
```

dbekaert:
> Note @bbuzz31 that I've encountered the following error in this test stack you shared that I didn't see before. Could you please advise? Thanks:

Sounds like the CRS was not embedded at the same level.

tools/ARIAtools/ARIAProduct.py
```diff
@@ -433,12 +433,21 @@ def __mappingVersion__(self, fname, version):
                 '/science/grids/imagingGeometry/azimuthAngle'
             ]
             if version.lower()=='1c':
+                # get weather model name
```
dbekaert: line 406. How is scene lenght used again? What are implications of use adding more overlap or making frames larger?

tools/ARIAtools/ARIAProduct.py
```diff
@@ -433,12 +433,21 @@ def __mappingVersion__(self, fname, version):
...
+                meta = gdal.Info(fname)
+                for i in meta.split():
+                    if '/science/grids/corrections/external/troposphere/' in i:
+                        model_name = i.split('/')[-3]
```
dbekaert: @sssangha Can you describe the behavior if two different models are in the product. e.g. product 1 has model A/B and product 2 has model B

tools/ARIAtools/ARIAProduct.py
```diff
@@ -433,12 +433,21 @@ def __mappingVersion__(self, fname, version):
...
+                        break
                 lyr_pref = '/science/grids/corrections'
                 sdskeys_addlyrs = [
                     lyr_pref + '/derived/ionosphere/ionosphere',
-                    lyr_pref + '/external/tides/solidEarth',
-                    lyr_pref + '/external/troposphere/troposphereWet',
-                    lyr_pref + '/external/troposphere/troposphereHydrostatic'
+                    lyr_pref + '/external/tides/solidEarth'
+                        '/reference/solidEarthTide',
+                    lyr_pref + f'/external/troposphere/{model_name}'
```
dbekaert: What is the expected working behavior under scenario if the solid earth tide layer is missing but tropo is present

tools/ARIAtools/ARIAProduct.py
sssangha: Yeah [this part](https://github.com/aria-tools/ARIA-tools/blob/sss_refsec/tools/ARIAtools/ARIAProduct.py#L447-#L449) of the code is intended to assign the weather models as separate keys for this very reason. I.e. as separate keys, they will be sorted as separately within the dictionary.

But it would be a good idea to confirm this is working as I'd intended. @bbuzz31, could you perhaps make a dummy-duplicate layer of HRES in the test files you've provided?

tools/ARIAtools/ARIAProduct.py
sssangha: So I simplified the existing check I had after I realized it didn't work as expected. Now the non-existence of such layers is captured and reported [here](https://github.com/aria-tools/ARIA-tools/blob/sss_refsec/tools/ARIAtools/ARIAProduct.py#L453) and [here](https://github.com/aria-tools/ARIA-tools/blob/sss_refsec/tools/ARIAtools/ARIAProduct.py#L477-#L486) in ARIAProduct. I confirmed it doesn't lead to issues in a test stack Brett provided which had tropo ref/sec, but no SET nor ionosphere.

Through initial sorting in ARIAProduct:
```
WARNING: Expected data layer key ionosphere not found in NETCDF:"/u/leffe-data/ssangha/solid_earth_test/temp_tropo_test/D071/products/S1-GUNW-D-R-071-tops-20200130_20200112-135156-34956N_32979N-PP-dd0c-v2_0_4.nc
WARNING: Expected data layer key solidEarthTide not found in NETCDF:"/u/leffe-data/ssangha/solid_earth_test/temp_tropo_test/D071/products/S1-GUNW-D-R-071-tops-20200130_20200112-135156-34956N_32979N-PP-dd0c-v2_0_4.nc
```

And again at TS stage:
```
WARNING: Stack for default ARIA TS layer ionosphere cannot be generated as it does not exist in any of the input products
WARNING: Stack for default ARIA TS layer solidEarthTide cannot be generated as it does not exist in any of the input products
```

tools/ARIAtools/ARIAProduct.py
bbuzzanga: I tested this with 4 products: all 4 had HRES, 2 also had GMAO. 
The HRES outputs are written correctly, and no GMAO outputs are calculated. I think this is expected behavior.

PR #351, opened 2023-03-08
Sequential unwrap sititcher 

mgovorcin:
adding new method for stitching frames of unwrapped phase along the track.

Sequential stitching is perform with forward and backward 2-pi integer cycle and range corrections using overlapping connected components between two neighboring frames


dbekaert (approved):
LGTM @mgovorcin thanks.

sssangha (approved):
LGTM!

PR #353, opened 2023-03-17
Handle correction layer offsets

sssangha:
Leverage @mgovorcin sequential stitching implementation for SET and tropo correction layers.

### DIFF
--- a/tools/ARIAtools/tsSetup.py
+++ b/tools/ARIAtools/tsSetup.py
@@ -95,13 +95,15 @@ def create_parser():
                         'threads for multiprocessing operation in gdal. '
                         'By default "2". Can also specify "All" to use all '
                         'available threads.')
-    # parser.add_argument('-sm', '--stitchMethod', dest='stitchMethodType',
-    #                      type=str, default='overlap', help='Method applied '
-    #                      'to stitch the unwrapped data. Either "overlap", '
-    #                      'where product overlap is minimized, or "2stage", '
-    #                      'where minimization is done on connected '
-    #                      'components, are allowed methods. '
-    #                      'default: "overlap".')
+    parser.add_argument('-sm', '--stitchMethod', dest='stitchMethodType',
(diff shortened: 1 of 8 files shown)


dbekaert:
per discussion; i beleive this is covering up an underlying problem in the SET code.

tools/ARIAtools/tsSetup.py
```diff
@@ -457,7 +461,7 @@ def main(inps=None):
         'mask': inps.mask,
         'outDir': inps.workdir,
         'outputFormat': inps.outputFormat,
-        'stitchMethodType': 'overlap',
+        'stitchMethodType': inps.stitchMethodType,
```
dbekaert, line 464: good catch

dbekaert (approved):
LGTM So looks like we are not supporting 3.8 anymore with the recent changes.

tools/ARIAtools/tsSetup.py
sssangha: Yeah, but tbf I think that was intentional before... I.e. for ariaTSsetup the whole `stitchMethodType` argument had been commented out to enforce overlap stitching as 2stage unwrapping was only other choice and experimental.

sssangha:
> LGTM So looks like we are not supporting 3.8 anymore with the recent changes.

Yup, which reminds me to update the readme! In summary, the new version of rioxarray utilizes tuples in a way that is no longer compatible with [python3.8](https://docs.python.org/3.8/library/typing.html#type-aliases) vs python 3.9 and [above](https://docs.python.org/3.9/library/typing.html#type-aliases) as @cmarshak discussed with me..

Something to track in our other repos!

sssangha:
@mgovorcin I cleaned up the sequential stitcher a little bit, namely making it inline with pep8 standard of a max of 79 characters per line + adjusting the variables in the metadata layer stitcher I introduced

mgovorcin (approved):

PR #356, opened 2023-03-31
Updates DEM from SRTM to Global Copernicus 90 meter DEM

cmarshak:
Uses this [PR](https://github.com/ACCESS-Cloud-Based-InSAR/dem-stitcher/pull/58) from dem-stitcher. Will convert it to "ready for review" from "draft" once we can download the version from conda. Removed the "open_topo" pieces.

### DIFF
--- a/tools/ARIAtools/mask_util.py
+++ b/tools/ARIAtools/mask_util.py
@@ -255,12 +255,12 @@ def arr2ds(ds_orig, arr, noData=0):
 
 
 class NLCDMasker(object):
-    def __init__(self, path_aria, lc=[11, 12, 90, 95]):
+    def __init__(self, path_aria, lc=None, dem_name='glo_90'):
         self.path_aria = path_aria
         self.path_bbox = op.join(self.path_aria, 'productBoundingBox',
                                                     'productBoundingBox.json')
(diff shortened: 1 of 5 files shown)


tools/ARIAtools/mask_util.py
```diff
@@ -255,12 +255,12 @@ def arr2ds(ds_orig, arr, noData=0):
 
 
 class NLCDMasker(object):
-    def __init__(self, path_aria, lc=[11, 12, 90, 95]):
+    def __init__(self, path_aria, lc=None, dem_name='glo_90'):
```
cmarshak: Changed the default argument because it will lead to horribly confusing errors.

https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

sssangha:
No longer expect existence of DEM in order to create a mask.

Also unrelated, removed `multithread = True` option in `gdal.Warp` commands accessing the standard products as newer gdal/HDF5 versions raise [numerous HDF5 errors](https://github.com/OSGeo/gdal/issues/7340).

cmarshak:
So - not totally sure about the changes you made, Sim. So I will defer to you on those.

sssangha (approved):
LGTM

Issue #352, opened 2023-03-16
[BUG] Coherence maps not multilooked


rzinke:
**Describe the bug**
`ariaTSsetup.py` does not properly multilook the coherence images.

**To Reproduce**
Applies to all data sets.

**Expected behavior**
When running `ariaTSsetup.py` with an argument passed to the multilook option, the `export_products` function should multilook all datasets. Instead, the coherence map does not appear to be downsampled or multilooked.

This is confirmed using `gdalinfo | grep "Pixel Size"` for a connected components file, which returns:
```
Pixel Size = (0.007501111605158,-0.007500000000000)
```

whereas doing the same for a coherence file returns:
```
Pixel Size = (0.002500000000000,-0.002500000000000)
```
(1/3 the dimension).

This leads to the behavior described in [MintPy Issue ticket 972](https://github.com/insarlab/MintPy/issues/972).

I will attempt to debug this week, time allowing.

rzinke:
Perhaps the issue is somewhere here:
https://github.com/aria-tools/ARIA-tools/blob/dev/tools/ARIAtools/extractProduct.py#L901-L907

As a test, I ran
```
ariaExtract.py -f "products/*.nc" -w 9arc -l coherence -ml 3
```
on a small dataset (`ariaDownload.py -t 121 -b "30.5 36.5 85.3 90.3" -s 20180101 -e 20180401`)

and received the following message:
```
...
Thread count specified for gdal multiprocessing = 2
/home/rzinke/miniconda3/envs/ARIA-tools/lib/python3.10/site-packages/shapely/set_operations.py:133: RuntimeWarning: invalid value encountered in intersection
  return lib.intersection(a, b, **kwargs)
Rejected scene 9arc/productBoundingBox/20180420_20180114.json has no common overlap with bbox
1 out of 29 interferograms rejected for not meeting specified spatial thresholds
Generating: coherence - [=>                       4%                       ] 20180102_20161126    0s /     0sHDF5-DIAG: Error detected in HDF5 (1.12.2) thread 1:
  #000: H5A.c line 528 in H5Aopen_by_name(): can't open attribute
    major: Attribute
    minor: Can't open object
  #001: H5VLcallback.c line 1091 in H5VL_attr_open(): attribute open failed
    major: Virtual Object Layer
    minor: Can't open object
  #002: H5VLcallback.c line 1058 in H5VL__attr_open(): attribute open failed
    major: Virtual Object Layer
    minor: Can't open object
[...]

Issue #359, opened 2023-05-03
[BUG] aztime error


rzinke:
**Describe the bug**
I am running `ariaTSsetup.py` to create a timeseries. Most of the products are computed, but an error is thrown before the final stack is generated. The error seems to indicate that an `aztime` cannot be found.

**To Reproduce**
1. Download 
```
ariaDownload.py -t 121 -b ../../AOIs/track_121_TibetB/track_121_TibetB.shp -s 20220101 -e 20221231
```

2. Command used
```
ariaTSsetup.py -f "products/*.nc" -w 9arc -b ../../AOIs/track_121_TibetB/track_121_TibetB.shp -d download -m download -nt 8 -mo 238000 -ml $looks -v
```

3. Error Output
```
...
NETCDF:"/data/Tibet/Track121/Timeseries/2022/products/S1-GUNW-D-R-121-tops-20220505_20220423-001148-00085E_00026N-PP-d008-v2_0_5.nc":/science/grids/data/unwrappedPhase is GDAL compatible
NETCDF:"/data/Tibet/Track121/Timeseries/2022/products/S1-GUNW-D-R-121-tops-20220505_20220423-001148-00085E_00026N-PP-d008-v2_0_5.nc":/science/grids/data/connectedComponents is GDAL compatible
NETCDF:"/data/Tibet/Track121/Timeseries/2022/products/S1-GUNW-D-R-121-tops-20220505_20220423-001213-00085E_00025N-PP-d2f3-v2_0_5.nc":/science/grids/data/unwrappedPhase is GDAL compatible
NETCDF:"/data/Tibet/Track121/Timeseries/2022/products/S1-GUNW-D-R-121-tops-20220505_20220423-001213-00085E_00025N-PP-d2f3-v2_0_5.nc":/science/grids/data/connectedComponents is GDAL compatible

Generating: coherence - [==================================================] 20220505_20220423  560s /    29s

Extracting single incidence angle, look angle and azimuth angle files valid over common interferometric grid
Generating: incidenceAngle - [==================================================] 20220117_20220105
Generating: lookAngle - [==================================================] 20220117_20220105
Generating: azimuthAngle - [==================================================] 20220117_20220105

Extracting perpendicular baseline grids for each interferogram pair
Generating: bPerpendicular - [==================================================] 20220505_20220423 1172s /    61s
All available tropo models are to be extracted
WARNING: User-requested tropo model HRES will not be generated as it does not exist in any of the input products
[...]

PR #360, opened 2023-05-09
Fix ML bug for unw/concomp fields

sssangha:
When the multilook option is specified and the user wishes to extract unw fields, the program had improperly applied the looks twice. This is because each call to `resampleRaster` (the multilook) function applies resampling to both unw and concomp fields in one go. However I caught two consecutive calls to `resampleRaster` passing unw and concomp, respectively. Thus, this is where "double-looks" were applied. This behavior has now been fixed, addressing issue #352

### DIFF
--- a/tools/ARIAtools.egg-info/PKG-INFO
+++ b/tools/ARIAtools.egg-info/PKG-INFO
@@ -1,5 +1,5 @@
 Metadata-Version: 2.1
 Name: ARIAtools
-Version: 1.1.4
+Version: 1.1.5
 Summary: This is the ARIA tools package without RelaxIV support
 License-File: LICENSE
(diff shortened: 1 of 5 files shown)


sssangha:
@rzinke I was unable to replicate the error you reported in #359 with the provided commands.

That said, please test it out again on your own end using this PR. Also make sure to update your env (e.g. `mamba env update --name ARIA-tools --file environment.yml --prune`). Not doing so will lead to a known annoyance in the form of a wall of benign warning messages (as seen in the print screen in #352). This issue + proposed solution have been documented and explained [elsewhere](https://stackoverflow.com/questions/42352841/how-to-update-an-existing-conda-environment-with-a-yml-file).

Also following our convo with @dbekaert I've pinpointed the issue regarding expected tropo layers. The ARIA class properly recognized these products as version `1b`, but the messages you saw in #359 concerning missing tropo layers appeared because originally the default behavior was to check for their existence regardless of product version. I've now adjusted the default argument for `-tm` to not make this [assumption](https://github.com/aria-tools/ARIA-tools/blob/54165442eddb3fe2824a08a30a0810ca9144e1c3/tools/ARIAtools/extractProduct.py#L72-L75), but to be clear this was a warning message that does not lead to crashes downstream.

rzinke (approved):
That fixed the issues. Thanks very much, @sssangha.

PR #362, opened 2023-05-11
Supress sequential stitcher print statements by default

sssangha:
Suppress the series of print statements made when running the sequential stitcher, even for verbose mode. They may be toggled manually by users wishing to debug the code, unless @mgovorcin advises otherwise?

Also, fix minor bug whereby program crashes if user specifies a `VRT` output. As with other layers, now default to `ENVI` in cases where user specifies `VRT`

### DIFF
--- a/tools/ARIAtools/sequential_stitching.py
+++ b/tools/ARIAtools/sequential_stitching.py
@@ -149,7 +149,8 @@ def stitch_2frames(unw_data1 : NDArray, conn_data1 : NDArray, rdict1 : dict,
                 conn_reverse[conn_reverse[:,0] == pair[0], 0] = \
                     np.float32(pair[1])
         else:
-            print('SKIPPED!:', correction, pair, '\n')
+            if verbose:
+                print('SKIPPED!:', correction, pair, '\n')
     
     # Backward correction
     for pair in conn_reverse:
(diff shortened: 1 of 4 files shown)


tools/ARIAtools/sequential_stitching.py
```diff
@@ -984,6 +985,10 @@ def write_GUNW_array(output_filename: Union[str, Path],
 
     array_type = gdal_array.NumericTypeCodeToGDALTypeCode(array.dtype)
 
+    # File must be physically extracted, cannot proceed with VRT format.
```
mgovorcin: if we leave this line here, then this function loses it modularity...we could just easily remove the option of selecting the format and set it every as ENVI. 

If this is needed for aria_extract, maybe it would be better to set it there to be  ENVI, aria_extract.py line 1135

mgovorcin: @sssangha  can you please add the closing figure/axes lines to the plotting function. 

fig.clear()
plt.close(fig)

Although, we should keep the plotting option  FALSE as default, as there still might be some memory leak https://github.com/matplotlib/matplotlib/pull/22002

mgovorcin (approved):
@sssangha  I added 2 small comments, other than that, looks good. Thanks

tools/ARIAtools/sequential_stitching.py
sssangha: @mgovorcin sounds good, I addressed your comments. I had to clean-up similar `outputformat` checks in other functions, but now this check is handled explicitly in aria_extract

mgovorcin: looks good

PR #369, opened 2023-07-21
bugfix

bbuzzanga:
think this was forgotten in posting update


mgovorcin (approved):
Thanks  @bbuzz31 for catching this one

PR #370, opened 2023-07-26
use the 'tap' on productplot coherence to have same size as watermask

### LINKED ISSUE #371: [BUG] Installation fails due to jupyter_contrib_nbextensions
conda/mamba cannot install `jupyter_contrib_nbextensions` into the ARIA-environment (python 3.9 and 3.10 tested)

The first of a bunch of no `ModuleNotFoundErrors`:
Executing transaction: - Traceback (most recent call last):
  File "/Users/buzzanga/mambaforge/lib/python3.10/site-packages/jupyter_contrib_core/notebook_compat/nbextensions.py", line 6, in <module>
    from notebook.extensions import BaseExtensionApp
ModuleNotFoundError: No module named 'notebook.extensions'

bbuzzanga:
there's a one pixel difference in size between watermask and coherence during plotting because the coherence is made with the new tap option like the watermask

UPDATE
Also use `pip` to install `jupyter_nb_extensions` as conda/mamba currently cannot. Closes #371

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -28,7 +28,7 @@ dependencies:
   - shapely
   #for ARIA-tools-docs
   - jupyterlab
-  - jupyter_contrib_nbextensions
   - dem_stitcher>=2.5.0
   - pip:
     - rise
+    - jupyter_contrib_nbextensions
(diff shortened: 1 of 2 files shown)


mgovorcin:
Thanks Brett

mgovorcin (approved):

PR #372, opened 2023-07-27
requirements: comment out dependencies for notebooks

yunjunz:
This PR comments out the notebook dependencies in the `requirements.txt` file, so one can easily install ARIA-tools and MintPy in one Python environment. 

Another perspective is that one may run `ARIA-tools` as python tools, without the need of running the notebooks in `ARIA-tools-docs`, having a `requirements.txt` file allows that.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -11,6 +11,7 @@ dependencies:
   - python>=3.8
   - asf_search
   - cartopy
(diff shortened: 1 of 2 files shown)


mgovorcin (approved):
LGTM thanks @yunjunz

PR #375, opened 2023-08-01
fig bug when flight direction specified

bbuzzanga:


tools/bin/ariaDownload.py
```diff
@@ -268,16 +268,23 @@ def query_asf(self):
         """Get the scenes from ASF"""
         bbox = make_bbox(self.inps.bbox)
         bbox = bbox.wkt if bbox is not None else None
+        if self.inps.flightdir == 'a':
+            flight_direction = 'ascending'
+        elif self.inps.flightdir == 'd':
+            flight_direction = 'descending'
+        else:
+            flight_direction = self.inps.flightdir
```
sssangha: @bbuzz31 was thinking of dummy-proofing as follow to deal with (workable) typos and case inconsistencies as so, let me know what you think:
```
        if self.inps.flightdir is not None:
            if self.inps.flightdir.lower()[0] == 'a':
                flight_direction = 'ascending'
            elif self.inps.flightdir == 'd':
                self.inps.flightdir.lower()[0]  = 'descending'
        else:
            flight_direction = None
```

bbuzzanga: LGTM!

sssangha (approved):
LGTM

Issue #364, opened 2023-05-26
[BUG] Inconsistent product dims


rzinke:
**Describe the bug**
I recently ordered some new (reprocessed) data for Tibet track 41. When I attempt to process those data for timeseries analysis using ARIA-tools, however, I get the error `Exception(f'Inconsistent product dims ...`. I understand that this is by design, as a check for consistent product dims was recently added; however, I'm not sure what to do about it.

Any suggestions?

***Full error output***
```
1164 out of 2936 interferograms rejected for not meeting specified spatial thresholds
Download/cropping DEM
Applied cutline to produce 3 arc-sec SRTM DEM: 9arc-no_v4/DEM/glo_90.dem
***Downloading water mask... ***

Extracting unwrapped phase, coherence, and connected components for each interferogram pair
Generating: unwrappedPhase - [>                                                 ]Traceback (most recent call last):
  File "/home/rzinke/miniconda3/envs/DevARIA-tools/bin/ariaTSsetup.py", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/rzinke/tools/ARIA-tools/tools/bin/ariaTSsetup.py", line 18, in <module>
    main(inps)
  File "/home/rzinke/tools/ARIA-tools/tools/ARIAtools/tsSetup.py", line 474, in main
    arrshape = export_products(standardproduct_info.products[1],
  File "/home/rzinke/tools/ARIA-tools/tools/ARIAtools/extractProduct.py", line 1164, in export_products
    raise Exception(f'Inconsistent product dims between'
Exception: Inconsistent product dims betweenproducts {outname} and {prev_outname}:check respective width ({ref_wid}, {ref_height})and height ({prod_wid}, {prod_height})
```

**Desktop (please complete the following information):**
 - ARIA-tools git tag: commit 54165442eddb3fe2824a08a30a0810ca9144e1c3 (HEAD -> sss_mlbugfix, origin/sss_mlbugfix)
 - OS: Linux CentOS

sssangha:
Hello @rzinke I have not encountered this error.

With that said, I've introduced a [PR](https://github.com/aria-tools/ARIA-tools/pull/363) with more robust checks + better handling of product resolution.

Could you please test it out for your case, and if this error persists could you perhaps provide a run example over a small stack I can debug? Thanks!

rzinke:
[...]

PR #377, opened 2023-08-06
Fix multilooking bug

sssangha:
Handle bug with multilooking by applying looks to the DEM and mask, before using for using or applying for product generation.

Addresses #364


rzinke:
Thanks, @sssangha. I am testing now. Will let you know asap.

rzinke (approved):
That fixed the cropping issue.

PR #378, opened 2023-08-08
Small fix of range correction bug

mgovorcin:
Description:
Range correction (mean phase shift from wrapped phase) sometimes gives opposite correction sign (skips the cycle) that affects the calculation of the shift between overlapping regions, e.g. instead of subtracting 3 radians it adds them up.

This happens in few cases, most of the time works well. I am not sure why this happens in the first place (could it be the ESD applied on one frame and another no?)

Solution:
I found it out that n integer of 2pi jumps with range correction most of the times is very close to the median difference. So if the difference between those two is larger than a PI, then I switch the sign of range correction. 

NOTE: still testing if the median difference between connected components is the most robust option for stitching to avoid the above mentioned.

Another changes:
Added check to accept dem and mask input as filename string instead of gdal.object to be able to use dask to run code in parallel.

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -1295,7 +1295,12 @@ def finalize_metadata(outname, bbox_bounds, dem_bounds, prods_TOTbbox, dem,
     from scipy.interpolate import RegularGridInterpolator
 
     # get final shape
-    arrres = gdal.Open(dem.GetDescription())
+    # MG: add option to pass dem path as string
+    if isinstance(dem, str):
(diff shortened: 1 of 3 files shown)


sssangha:
Hello @mgovorcin this LGTM. The velocity field derived from the pairs spanning SF bay through 2019 (the subject of the ariaTSsetup notebook) looks good:
Screenshot 2023-08-09 at 11 38 24 AM

I made the sequential stitcher the default. Let me know if there are any further changes you anticipate for now. Otherwise I will merge and thereafter issue a new release.

sssangha (approved):
LGTM

PR #381, opened 2023-09-06
Sort pairs by latitude

sssangha:
The dictionary is sorted by `azimuthZeroDopplerMidTime` (see [here](https://github.com/aria-tools/ARIA-tools/blob/dev/tools/ARIAtools/ARIAProduct.py#L756)), which is derived from just the [filename](https://github.com/aria-tools/ARIA-tools/blob/dev/tools/ARIAtools/ARIAProduct.py#L395) to save time from cracking the product open.

But in some cases it’s the same value for adjacent frames (e.g. `S1-GUNW-A-R-033-tops-20170217_20150815-225114-00076W_00040N-PP-413c-v3_0_0.nc` and `S1-GUNW-A-R-033-tops-20170217_20150815-225114-00076W_00041N-PP-d372-v3_0_0.nc`) and the frames are then flipped around in sorting, leading to a false positive in that the program rejects these pairs due to an apparent gap.

@bbuzz31 you can confirm this patch now avoids this behavior by appropriately passing all of the "2017" pairs in the subdirectory you shared by running the following command:
> ariaExtract.py -f '/u/leffe-data2/buzzanga/data/VLM/Sentinel1/EastCoast/track_033/products_track033/*2017*.nc' -of ISCE -b ../bbox_selected.GeoJSON -verbose -w test_all

### DIFF
--- a/tools/ARIAtools/ARIAProduct.py
+++ b/tools/ARIAtools/ARIAProduct.py
@@ -398,6 +398,16 @@ class ARIA_standardproduct:
                 + basename.split('-')[7][2:4] + ':' \
                 + basename.split('-')[7][4:] + '.0'
 
+            # assign latitude to assist with sorting
+            rdrmetadata_dict['centerLatitude'] =  \
+                basename.split('-')[8].split('_')[1]
+            if rdrmetadata_dict['centerLatitude'][-1] == 'S':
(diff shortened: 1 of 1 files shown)


bbuzzanga (approved):
LGTM

PR #390, opened 2024-03-20
Add NISAR product support

sssangha:


tools/ARIAtools/ARIAProduct.py
```diff
@@ -527,6 +585,166 @@ class (see instructions on how to appropriately add new
...
+            pyproj_transformer.transform(x, y), latlon_file_bbox)
+
+        # hardcoded keys
+        rdrmetadata_dict['missionID'] = 'NISAR'
+        rdrmetadata_dict['productType'] = 'UNW GEO IFG'
+        rdrmetadata_dict['wavelength'] = 0.24
+        rdrmetadata_dict['centerFrequency'] = 1.2699997500604727e+09
+        rdrmetadata_dict['slantRangeSpacing'] = 4.461197291666666
+        rdrmetadata_dict['slantRangeStart'] = 727415.98682514
+        rdrmetadata_dict['slantRangeEnd'] = 791384.81843777
```
hfattahi: Any reason to hardcode any of these parameters? They all exist in the NISAR GUNW products. The wavelength and center frequency may be different for different modes and should not be assumed constant for all NISAR L-SAR modes.

sssangha: @hfattahi yes, those hardcoded values were a vestige. I went ahead and accessed the appropriate parameters directly.

PR #392, opened 2024-04-03
Prioritize v3 over v2 products

sssangha:
For a given IFG, avoid mixing v3 and v2 products such that only v3 products are passed for a given IFG where available.

### DIFF
--- a/tools/ARIAtools/product.py
+++ b/tools/ARIAtools/product.py
@@ -128,45 +128,50 @@ def remove_scenes(products):
 
                 # Check if IFG dict corresponding to ref prod already exists
                 # and if it does then append values
-                try:
-                    dict_ind = sorted_products.index(next(
-                        item for item in sorted_products
-                        if scene[1]['productBoundingBox'] in
-                        item[1]['productBoundingBox']))
+                dict_item = None
+                for item in sorted_products:
+                    if scene[1]['productBoundingBox'] in \
+                       item[1]['productBoundingBox']:
+                        dict_item = item
+                        break
+                if dict_item is not None:
+                    dict_ind = sorted_products.index(dict_item)
                     dict_1 = package_dict(
                         scene, new_scene, 0, sorted_products, dict_ind)
                     dict_2 = package_dict(
                         scene, new_scene, 1, sorted_products, dict_ind)
                     sorted_products[dict_ind] = [dict_1, dict_2]
-
                 # Match IFG corresponding to reference product NOT found
                 # so initialize dictionary for new IFG
-                except BaseException:
+                else:
                     dict_1 = package_dict(scene, new_scene, 0)
                     dict_2 = package_dict(scene, new_scene, 1)
                     new_dict = [dict_1, dict_2]
                     sorted_products.extend([new_dict])
+
             # If prods correspond to different orbits entirely
             else:
+
                 # Check if IFG dict corresponding to ref prod already exists
                 # and if it does not then pass as new IFG
-                # TODO clean this conditional
-                if [item for item in sorted_products
-                    if scene[1]['productBoundingBox'] in
-                    item[1]['productBoundingBox']] == []:
-
+                track_existing_ifg = []
+                for item in sorted_products:
+                    if scene[1]['productBoundingBox'] in \
+                    item[1]['productBoundingBox']:
+                        track_existing_ifg.append(item)
+                if track_existing_ifg == []:
                     dict_1 = package_dict(scene, scene, 0)
                     dict_2 = package_dict(scene, scene, 1)
                     new_dict = [dict_1, dict_2]
                     sorted_products.extend([new_dict])
-
                 # Check if IFG dict corresponding to ref prod already exists
                 # and if it does not then pass as new IFG
-                # TODO clean this conditional
-                if [item for item in sorted_products
-                        if new_scene[1]['productBoundingBox'] in
-                    item[1]['productBoundingBox']] == []:
-
+                track_existing_ifg = []
+                for item in sorted_products:
+                    if new_scene[1]['productBoundingBox'] in \
+                    item[1]['productBoundingBox']:
+                        track_existing_ifg.append(item)
+                if track_existing_ifg == []:
                     dict_1 = package_dict(new_scene, new_scene, 0)
                     dict_2 = package_dict(new_scene, new_scene, 1)
                     new_dict = [dict_1, dict_2]
@@ -334,7 +339,7 @@ class Product:
 
                 try:
                     bbox = [float(val) for val in bbox.split()]
-                except BaseException:
+                except ValueError:
                     raise Exception(
                         'Cannot understand the --bbox argument. String input '
                         'is incorrect or path does not exist.')
@@ -376,13 +381,13 @@ class Product:
         number, and populate corresponding product dictionary accordingly.
         """
         # Get standard product version from file
-        try:
-            # version accessed differently between URL vs downloaded product
-            version = str(
-                osgeo.gdal.Open(fname).GetMetadataItem('NC_GLOBAL#version'))
-
-        except BaseException:
-            LOGGER.warning('%s is not a supported file type... skipping', fname)
+        # version accessed differently between URL vs downloaded product
+        version = str(
+            osgeo.gdal.Open(fname).GetMetadataItem('NC_GLOBAL#version'))
+        if version == 'None':
+            LOGGER.warning(
+                f'{fname} is not a supported file '
+                'type... skipping')
             return []
 
         # Enforce forward-compatibility of netcdf versions
@@ -818,20 +823,22 @@ class Product:
 
                 # Check if IFG dict corresponding to ref prod already exists
                 # and if it does then append values
-                try:
-                    dict_ind = sorted_products.index(next(
-                        item for item in sorted_products
-                        if scene[1]['productBoundingBox'] in
-                        item[1]['productBoundingBox']))
+                dict_item = None
+                for item in sorted_products:
+                    if scene[1]['productBoundingBox'] in \
+                       item[1]['productBoundingBox']:
+                        dict_item = item
+                        break
+                if dict_item is not None:
+                    dict_ind = sorted_products.index(dict_item)
                     dict_1 = package_dict(
                         scene, new_scene, 0, sorted_products, dict_ind)
                     dict_2 = package_dict(
                         scene, new_scene, 1, sorted_products, dict_ind)
                     sorted_products[dict_ind] = [dict_1, dict_2]
-
                 # Match IFG corresponding to reference product NOT found
                 # so initialize dictionary for new IFG
-                except BaseException:
+                else:
                     dict_1 = package_dict(scene, new_scene, 0)
                     dict_2 = package_dict(scene, new_scene, 1)
                     new_dict = [dict_1, dict_2]
@@ -851,29 +858,29 @@ class Product:
 
             # If prods correspond to different orbits entirely
             else:
+
                 # Check if IFG dict corresponding to ref prod already exists
                 # and if it does not then pass as new IFG
-                # TODO clean this conditional
-                if [item for item in sorted_products
-                    if scene[1]['productBoundingBox'] in
-                    item[1]['productBoundingBox']] == [] \
-                    and scene[0]['pair_name'] not in \
-                        track_rejected_pairs:
-
+                track_existing_ifg = []
+                for item in sorted_products:
+                    if scene[1]['productBoundingBox'] in \
+                    item[1]['productBoundingBox']:
+                        track_existing_ifg.append(item)
+                if track_existing_ifg == [] and \
+                    scene[0]['pair_name'] not in track_rejected_pairs:
                     dict_1 = package_dict(scene, scene, 0)
                     dict_2 = package_dict(scene, scene, 1)
                     new_dict = [dict_1, dict_2]
                     sorted_products.extend([new_dict])
-
                 # Check if IFG dict corresponding to ref prod already exists
                 # and if it does not then pass as new IFG
-                # TODO clean this conditional
-                if [item for item in sorted_products
-                        if new_scene[1]['productBoundingBox'] in
-                    item[1]['productBoundingBox']] == [] \
-                        and new_scene[0]['pair_name'] not in \
-                    track_rejected_pairs:
-
+                track_existing_ifg = []
+                for item in sorted_products:
+                    if new_scene[1]['productBoundingBox'] in \
+                    item[1]['productBoundingBox']:
+                        track_existing_ifg.append(item)
+                if track_existing_ifg == [] and \
+                    new_scene[0]['pair_name'] not in track_rejected_pairs:
                     dict_1 = package_dict(new_scene, new_scene, 0)
                     dict_2 = package_dict(new_scene, new_scene, 1)
                     new_dict = [dict_1, dict_2]
--- a/tools/ARIAtools/product.py
+++ b/tools/ARIAtools/product.py
@@ -96,6 +96,127 @@ def package_dict(scene, new_scene, scene_ind,
     return new_dict
 
 
+def remove_scenes(products):
+    """
+    For ARIA-S1 GUNW products, identify and reject legacy v2 products if
+    in the presence of v3 products for a given IFG.
+
+    'products' = list of input products
+    """
+    track_legacy_products = []
+    sorted_products = []
+
+    # only check ARIA-S1 GUNW products
+    if os.path.basename(products[0][1]['productBoundingBox'].split( \
+        '"')[1])[:2] == 'S1':
+        for i in enumerate(products[:-1]):
+            scene = i[1]
+            new_scene = products[i[0] + 1]
+            scene_t_ref = datetime.datetime.strptime(
+                scene[0]['pair_name'][:8], "%Y%m%d")
+            new_scene_t_ref = datetime.datetime.strptime(
+                new_scene[0]['pair_name'][:8], "%Y%m%d")
+            scene_t = datetime.datetime.strptime(
+                scene[0]['pair_name'][9:], "%Y%m%d")
+            new_scene_t = datetime.datetime.strptime(
+                new_scene[0]['pair_name'][9:], "%Y%m%d")
+
+            # Only pass scene if it temporally (i.e. in same orbit)
+            # overlaps with reference scene
+            if abs(new_scene_t - scene_t) <= ONE_DAY and \
+                abs(new_scene_t_ref - scene_t_ref) <= ONE_DAY:
+                print('inside')
+
+                # Check if IFG dict corresponding to ref prod already exists
+                # and if it does then append values
+                try:
+                    dict_ind = sorted_products.index(next(
+                        item for item in sorted_products
+                        if scene[1]['productBoundingBox'] in
+                        item[1]['productBoundingBox']))
+                    dict_1 = package_dict(
+                        scene, new_scene, 0, sorted_products, dict_ind)
+                    dict_2 = package_dict(
+                        scene, new_scene, 1, sorted_products, dict_ind)
+                    sorted_products[dict_ind] = [dict_1, dict_2]
+
+                # Match IFG corresponding to reference product NOT found
+                # so initialize dictionary for new IFG
+                except BaseException:
+                    dict_1 = package_dict(scene, new_scene, 0)
+                    dict_2 = package_dict(scene, new_scene, 1)
+                    new_dict = [dict_1, dict_2]
+                    sorted_products.extend([new_dict])
+            # If prods correspond to different orbits entirely
+            else:
+                # Check if IFG dict corresponding to ref prod already exists
+                # and if it does not then pass as new IFG
+                # TODO clean this conditional
+                if [item for item in sorted_products
+                    if scene[1]['productBoundingBox'] in
+                    item[1]['productBoundingBox']] == []:
+
+                    dict_1 = package_dict(scene, scene, 0)
+                    dict_2 = package_dict(scene, scene, 1)
+                    new_dict = [dict_1, dict_2]
+                    sorted_products.extend([new_dict])
+
+                # Check if IFG dict corresponding to ref prod already exists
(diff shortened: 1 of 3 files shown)


sssangha:
@alexfore , I made some changes to first filter out the v2 products before checking for gaps. I wont make any further commits until and unless you advice me to. Thanks

Edit: Just made one small commit to remove a debugging print statement!

alexfore:
This is probably fine. My concern atm is about the env we run this software in

tools/ARIAtools/product.py
```diff
@@ -74,6 +96,126 @@ def package_dict(scene, new_scene, scene_ind,
...
+                        item[1]['productBoundingBox']))
+                    dict_1 = package_dict(
+                        scene, new_scene, 0, sorted_products, dict_ind)
+                    dict_2 = package_dict(
+                        scene, new_scene, 1, sorted_products, dict_ind)
+                    sorted_products[dict_ind] = [dict_1, dict_2]
+
+                # Match IFG corresponding to reference product NOT found
+                # so initialize dictionary for new IFG
+                except BaseException:
```
alexfore: What exception is expected to be raised here, can we use that instead?

tools/ARIAtools/product.py
```diff
@@ -74,6 +96,126 @@ def package_dict(scene, new_scene, scene_ind,
...
+
+            # Only pass scene if it temporally (i.e. in same orbit)
+            # overlaps with reference scene
+            if abs(new_scene_t - scene_t) <= ONE_DAY and \
+                abs(new_scene_t_ref - scene_t_ref) <= ONE_DAY:
+
+                # Check if IFG dict corresponding to ref prod already exists
+                # and if it does then append values
+                try:
+                    dict_ind = sorted_products.index(next(
```
alexfore: IMO this list comp is very confusing I can't tell what is being done here without significant mental effort. same comment on other list comps below

tools/ARIAtools/product.py
```diff
@@ -74,6 +96,126 @@ def package_dict(scene, new_scene, scene_ind,
...
+                    dict_1 = package_dict(scene, new_scene, 0)
+                    dict_2 = package_dict(scene, new_scene, 1)
+                    new_dict = [dict_1, dict_2]
+                    sorted_products.extend([new_dict])
+            # If prods correspond to different orbits entirely
+            else:
+                # Check if IFG dict corresponding to ref prod already exists
+                # and if it does not then pass as new IFG
+                # TODO clean this conditional
+                if [item for item in sorted_products
```
alexfore: same as above re list comps

tools/ARIAtools/product.py
```diff
@@ -74,6 +96,126 @@ def package_dict(scene, new_scene, scene_ind,
...
+
+                    dict_1 = package_dict(scene, scene, 0)
+                    dict_2 = package_dict(scene, scene, 1)
+                    new_dict = [dict_1, dict_2]
+                    sorted_products.extend([new_dict])
+
+                # Check if IFG dict corresponding to ref prod already exists
+                # and if it does not then pass as new IFG
+                # TODO clean this conditional
+                if [item for item in sorted_products
```
alexfore: same as above re list comps

tools/ARIAtools/product.py
bbuzzanga: john-mclane

tools/ARIAtools/product.py
sssangha: Ok, went ahead and updated. When removing the list comp though, the need for using the exception is vestigial. I replaced with simple if/else logic

sssangha:
@alexfore I went ahead and replaced the list comps you pointed to into hopefully more appropriate and digestible `for` loops. Please let me know what you think. Thanks

alexfore:
I think a lot of the way the software interfaces / uses the `products` variable is confusing and inside-out.  Meaning that many parts of the software perform complicated conditionals (list comps, or how those were rewritten in here) to verify if `products` has or does not have some string or item in one of its many parts. All of these pieces of code have to know how to deal with this variable to perform various tests, add or remove things from it, ...etc.  This adds complexity and maintenance issues (for example it is not clear what is being tested or how to modify it to new requirements). I strongly feel code should be as simple as possible and the syntax as clear and direct as possible.

I'm thinking of some ways to address that, for example we could subclass `dict` and add some methods to that class to encapsulate these confusing bits so that the calling code can have simple and clear syntax. Then only the class needs to understand the data design of that `products` variable / schema and we can contain that complexity rather than letting it spill out to all these other bits of code.

So the above is clearly out of scope of this PR, so maybe this is something I should keep thinking about as future improvements / refactoring.

For this PR I think I'd like to see some more formatting changes to make it pep8, specifically with line continuation formatting (implied line continuation using parens, brackets, braces is prefered over `\`), and alignment of those continued lines)

I've also read that using f-strings or `format` in any logging is generally discouraged over c-style formatting / printf formatting,

sssangha:
Sounds good @alexfore , we should definitely think of a new, simpler alternative to the way the `products` variable is passed perhaps in another PR.

As far are pep8 changes, is this something we should perhaps also target in another PR? I'm concerned about making additional changes beyond the scope of what I had initially even touched here since it would diverge from the other [branch](https://github.com/aria-tools/ARIA-tools/pull/392) I pointed to with NISAR support we also want to merge.

bbuzzanga:
> Sounds good @alexfore , we should definitely think of a new, simpler alternative to the way the `products` variable is passed perhaps in another PR.
> 
> As far are pep8 changes, is this something we should perhaps also target in another PR? I'm concerned about making additional changes beyond the scope of what I had initially even touched here since it would diverge from the other [branch](https://github.com/aria-tools/ARIA-tools/pull/392) I pointed to with NISAR support we also want to merge.

+1

alexfore:
> As far are pep8 changes, is this something we should perhaps also target in another PR? I'm concerned about making additional changes beyond the scope of what I had initially even touched here since it would diverge from the other [branch](https://github.com/aria-tools/ARIA-tools/pull/392) I pointed to with NISAR support we also want to merge.

I only mean those lines of code which are contained in this PR, most of the conditionals/guards which use line continuation use `\` and/or have mis-aligned indents (not pep8)

I should have already addressed most of the pep8 issuse already in the dev_refactor branch, I ran it through an auto formatter

sssangha:
Ok sounds good. Will make the appropriate adjustments to the changes I've introduced here and in the other NISAR support branch.

sssangha:
Hello @alexfore , ok I went ahead and made the appropriate changes through an auto pep8 formatter. I also made sure logging calls no longer use f-strings.

I'm working now to go through and make the appropriate changes for the NISAR [branch](https://github.com/aria-tools/ARIA-tools/pull/392).

tools/ARIAtools/product.py
```diff
@@ -811,7 +976,7 @@ def __run__(self):
             LOGGER.debug([
```
alexfore: logger should be fed basic strings or something to be lazy formatted and not lists
last line not aligned correctly

tools/ARIAtools/product.py
```diff
@@ -706,29 +867,29 @@ def __continuous_time__(self):
...
-                        and new_scene[0]['pair_name'] not in \
-                    track_rejected_pairs:
-
+                track_existing_ifg = []
+                for item in sorted_products:
+                    if new_scene[1]['productBoundingBox'] in \
+                            item[1]['productBoundingBox']:
+                        track_existing_ifg.append(item)
+                if track_existing_ifg == [] and \
+                        new_scene[0]['pair_name'] not in track_rejected_pairs:
```
alexfore: prefer parens to `\` for line continuation 
second line not aligned correctly

tools/ARIAtools/product.py
```diff
@@ -706,29 +867,29 @@ def __continuous_time__(self):
...
-                if [item for item in sorted_products
-                        if new_scene[1]['productBoundingBox'] in
-                    item[1]['productBoundingBox']] == [] \
-                        and new_scene[0]['pair_name'] not in \
-                    track_rejected_pairs:
-
+                track_existing_ifg = []
+                for item in sorted_products:
+                    if new_scene[1]['productBoundingBox'] in \
+                            item[1]['productBoundingBox']:
```
alexfore: prefer parens to `\` for line continuation 
second line not aligned correctly

tools/ARIAtools/product.py
```diff
@@ -706,29 +867,29 @@ def __continuous_time__(self):
...
-                    and scene[0]['pair_name'] not in \
-                        track_rejected_pairs:
-
+                track_existing_ifg = []
+                for item in sorted_products:
+                    if scene[1]['productBoundingBox'] in \
+                            item[1]['productBoundingBox']:
+                        track_existing_ifg.append(item)
+                if track_existing_ifg == [] and \
+                        scene[0]['pair_name'] not in track_rejected_pairs:
```
alexfore: prefer parens to `\` for line continuation 
second line not aligned correctly

tools/ARIAtools/product.py
```diff
@@ -706,29 +867,29 @@ def __continuous_time__(self):
...
-                if [item for item in sorted_products
-                    if scene[1]['productBoundingBox'] in
-                    item[1]['productBoundingBox']] == [] \
-                    and scene[0]['pair_name'] not in \
-                        track_rejected_pairs:
-
+                track_existing_ifg = []
+                for item in sorted_products:
+                    if scene[1]['productBoundingBox'] in \
+                            item[1]['productBoundingBox']:
```
alexfore: prefer parens to `\` for line continuation 
second line not aligned correctly

tools/ARIAtools/product.py
```diff
@@ -673,20 +832,22 @@ def __continuous_time__(self):
...
                 # and if it does then append values
-                try:
-                    dict_ind = sorted_products.index(next(
-                        item for item in sorted_products
-                        if scene[1]['productBoundingBox'] in
-                        item[1]['productBoundingBox']))
+                dict_item = None
+                for item in sorted_products:
+                    if scene[1]['productBoundingBox'] in \
+                       item[1]['productBoundingBox']:
```
alexfore: prefer parens to `\` for line continuation

tools/ARIAtools/product.py
```diff
@@ -662,7 +821,7 @@ def __continuous_time__(self):
             # and spatially overlaps with reference scene
             if (scene_area.intersection(new_scene_area).area > 0. and
                 abs(new_scene_t - scene_t) <= ONE_DAY and
-                abs(new_scene_t_ref - scene_t_ref) <= ONE_DAY):
+                    abs(new_scene_t_ref - scene_t_ref) <= ONE_DAY):
```
alexfore: last line not aligned correctly

tools/ARIAtools/product.py
```diff
@@ -541,15 +698,14 @@ class (see instructions on how to appropriately add new
...
         # Setup datalyr_dict
         datalyr_dict = {}
-        for i in enumerate(these_layer_keys):
-            datalyr_dict[i[1]] = fname + '":' + sdskeys[i[0]]
         datalyr_dict['pair_name'] = self.pairname
-
         # 'productBoundingBox' will be updated to point to shapefile
         # corresponding to final output raster, so record of
         # indivdual frames preserved here
         datalyr_dict['productBoundingBoxFrames'] = \
```
alexfore: prefer parens to `\` for line continuation

tools/ARIAtools/product.py
```diff
@@ -44,42 +46,197 @@ def package_dict(scene, new_scene, scene_ind,
...
+            if v2_prods and v3_prods:
+                legacy_indices = [ind for ind, val in enumerate(vers)
+                                  if val < 300.]
+                track_legacy_products.extend([val for ind, val in enumerate(
+                    i[1]['unwrappedPhase']) if ind in legacy_indices])
+
+                # Iterate over each dictionary to remove these keys
+                LOGGER.debug('The following v2 products were rejected '
+                             'to ensure they are not mixed with v3 products:')
+                LOGGER.debug([os.path.basename(
```
alexfore: logger should be fed a string or something which can be lazy formatted

tools/ARIAtools/product.py
```diff
@@ -44,42 +46,197 @@ def package_dict(scene, new_scene, scene_ind,
...
+                    dict_1 = package_dict(scene, scene, 0)
+                    dict_2 = package_dict(scene, scene, 1)
+                    new_dict = [dict_1, dict_2]
+                    sorted_products.extend([new_dict])
+                # Check if IFG dict corresponding to ref prod already exists
+                # and if it does not then pass as new IFG
+                track_existing_ifg = []
+                for item in sorted_products:
+                    if new_scene[1]['productBoundingBox'] in \
+                            item[1]['productBoundingBox']:
```
alexfore: recommend use parens here and not `\`
continued line is not aligned correctly

tools/ARIAtools/product.py
```diff
@@ -44,42 +46,197 @@ def package_dict(scene, new_scene, scene_ind,
...
+
+            # If prods correspond to different orbits entirely
+            else:
+
+                # Check if IFG dict corresponding to ref prod already exists
+                # and if it does not then pass as new IFG
+                track_existing_ifg = []
+                for item in sorted_products:
+                    if scene[1]['productBoundingBox'] in \
+                            item[1]['productBoundingBox']:
```
alexfore: recommend use parens here and not `\`
continued line is not aligned correctly

tools/ARIAtools/product.py
```diff
@@ -44,42 +46,197 @@ def package_dict(scene, new_scene, scene_ind,
...
+            # Only pass scene if it temporally (i.e. in same orbit)
+            # overlaps with reference scene
+            if abs(new_scene_t - scene_t) <= ONE_DAY and \
+                    abs(new_scene_t_ref - scene_t_ref) <= ONE_DAY:
+
+                # Check if IFG dict corresponding to ref prod already exists
+                # and if it does then append values
+                dict_item = None
+                for item in sorted_products:
+                    if scene[1]['productBoundingBox'] in \
```
alexfore: recommend use parens here and not `\`

tools/ARIAtools/product.py
```diff
@@ -44,42 +46,197 @@ def package_dict(scene, new_scene, scene_ind,
...
+            new_scene_t_ref = datetime.datetime.strptime(
+                new_scene[0]['pair_name'][:8], "%Y%m%d")
+            scene_t = datetime.datetime.strptime(
+                scene[0]['pair_name'][9:], "%Y%m%d")
+            new_scene_t = datetime.datetime.strptime(
+                new_scene[0]['pair_name'][9:], "%Y%m%d")
+
+            # Only pass scene if it temporally (i.e. in same orbit)
+            # overlaps with reference scene
+            if abs(new_scene_t - scene_t) <= ONE_DAY and \
```
alexfore: recommend use parens here and not `\`
continued line is not aligned correctly

tools/ARIAtools/product.py
```diff
@@ -44,42 +46,197 @@ def package_dict(scene, new_scene, scene_ind,
...
-            subitem for item in a for subitem in (item if
-            isinstance(item, list) else [item])] for a in zip(
+        # first pass empty list if key does not exist for a scene
+        for i in dict_keys:
+            if i not in dict_keys_ref:
+                sorted_dict[dict_ind][scene_ind][i] = []
+            if i not in dict_keys_sec:
+                new_scene[scene_ind][i] = []
+        # then merge
+        dict_vals = [
```
alexfore: IMO this is worse than before

tools/ARIAtools/product.py
sssangha: This part seemed to be inherited in the original refactoring effort (i.e. I didn't touch it). I will adjust as appropriate though.

tools/ARIAtools/product.py
sssangha: Apologies, the autopep8 tool I use keeps overriding this. Will fix it.

sssangha:
Hello @alexfore, I went ahead and worked to incorporate the pep8 formatting tweaks to my commits.

However, I've reverted the one list comprehension since I was having a hard time to iterate off of it and it was part of the legacy code that I didn't touch with these commits. @mgovorcin changes on his [branch](https://github.com/aria-tools/ARIA-tools/blob/dask_parallelize/tools/ARIAtools/contrib/ARIA_product.py) address this more, and I think it would be best to swap out the list comprehensions through another follow-up PR more focused on his proposed changes.

So that said, please let me know if this particular PR is ready. I'll continue to work to clean up the commits I made to the other branch `dev_refactor_nisar`. Thanks!

alexfore (approved):
Thanks @sssangha

dbekaert:
@bbuzz31 can you test the logic of V2-V3?

bbuzzanga:
I'm getting a bunch of warnings which may be okay, but then an error which looks unpleasant regarding size inconsistencies.

`ariaExtract.py -f "products/*" -l 'all' -w BB_default -d Download` # using test products from v2v3 test dataset

```
/Users/buzzanga/mambaforge/envs/ARIA-tools/bin/ariaExtract.py:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__('pkg_resources').require('ARIAtools==1.1.6')
2024-04-11 08:24:52,753 - ARIAtools.util.vrt - WARNING - Extraction of raider-derived troposphere layers is not possible as specified tropo model name(s) "-tm None" is not valid.
2024-04-11 08:24:53,504 - ARIAtools.util.dem - WARNING - Cannot proceed with VRT format, using ENVI format instead
Downloading glo_90 tiles: 100%|████████████████████████████████████████████████████████████████████████| 20/20 [01:47<00:00,  5.38s/it]
2024-04-11 08:26:44,368 - rasterio._env - WARNING - CPLE_NotSupported in driver ENVI does not support creation option BLOCKYSIZE
2024-04-11 08:26:44,441 - rasterio._env - WARNING - CPLE_NotSupported in driver ENVI does not support creation option BLOCKYSIZE
Generating: solidEarthTide - [===============>        33%                       ] 20220331_20220319    0s /     1s2024-04-11 08:26:44,889 - rasterio._env - WARNING - CPLE_NotSupported in driver ENVI does not support creation option BLOCKYSIZE
2024-04-11 08:26:44,962 - rasterio._env - WARNING - CPLE_NotSupported in driver ENVI does not support creation option BLOCKYSIZE
Generating: solidEarthTide - [======================= 67% =====>                ] 20230829_20230724    1s /     0s2024-04-11 08:26:45,423 - rasterio._env - WARNING - CPLE_NotSupported in driver ENVI does not support creation option BLOCKYSIZE
2024-04-11 08:26:45,491 - rasterio._env - WARNING - CPLE_NotSupported in driver ENVI does not support creation option BLOCKYSIZE
Generating: solidEarthTide - [===============>        33%                       ] 20230829_20230817    1s /     3s
/Users/buzzanga/Software_InSAR/ARIA-tools_git/tools/ARIAtools/util/stitch.py:369: RuntimeWarning: invalid value encountered in multiply
  comb_data = np.empty((n, length, width), dtype=np.float64) * np.nan
  File "/Users/buzzanga/mambaforge/envs/ARIA-tools/bin/ariaExtract.py", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/buzzanga/Software_InSAR/ARIA-tools_git/tools/bin/ariaExtract.py", line 283, in <module>
    main()
  File "/Users/buzzanga/Software_InSAR/ARIA-tools_git/tools/bin/ariaExtract.py", line 270, in main
    arrshape = ARIAtools.extractProduct.export_products(**export_dict)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/buzzanga/Software_InSAR/ARIA-tools_git/tools/ARIAtools/extractProduct.py", line 1221, in export_products
    ARIAtools.util.vrt.dim_check(ref_arr, prod_arr)
  File "/Users/buzzanga/Software_InSAR/ARIA-tools_git/tools/ARIAtools/util/vrt.py", line 575, in dim_check
    raise Exception(
Exception: Inconsistent product dims between products BB_default/lookAngle/20210330_20200329 and BB_default/unwrappedPhase/20210330_20200329: check respective width (4545, 4379) and height (3821, 3382) and geotrans ((-123.066666666667, 0.000802896956362301, -0.0, 43.0408333333333, -0.0, -0.000737590508592863), (-123.066666666667, 0.000833333333333333, -0.0, 43.0408333333333, -0.0, -0.000833333333333333))
```

sssangha:
@bbuzz31 , I was able to replicate this bug. It looks like it wasn't associated with the changes I introduced here. However, I did track down the culprit.

The expanded, uncropped DEM (i.e. the DEM without the cutline applied) is needed to successfully intersect and interpolate the metadata fields. However, I needed to adjust the logic of `finalize_metadata` to appropriately pass and enforce consistent bounds and spacing in the final fields vis-a-vis the other full resolution layers. I can now verify the patch fixes the bug.

I think the rasterio warning is a benign issue. Perhaps would be better to take care of though in a follow-up PR.

bbuzzanga:
Hey @sssangha thanks works great. One thing maybe to flag is that if you try and reuse an output directory the issue comes back. If I recall, we had other issues when reusing directories so maybe this can be addressed with larger refactoring effort

sssangha:
> Hey @sssangha thanks works great. One thing maybe to flag is that if you try and reuse an output directory the issue comes back. If I recall, we had other issues when reusing directories so maybe this can be addressed with larger refactoring effort

Yeah that's correct. What you brought up would be relevant to a broader, more involved dedup effort. In the meantime, the user should specify a fresh output directory or just delete an existing output directory.

bbuzzanga:
> > Hey @sssangha thanks works great. One thing maybe to flag is that if you try and reuse an output directory the issue comes back. If I recall, we had other issues when reusing directories so maybe this can be addressed with larger refactoring effort
> 
> Yeah that's correct. What you brought up would be relevant to a broader, more involved dedup effort. In the meantime, the user should specify a fresh output directory or just delete an existing output directory.

Kk. Might be worth noting in the readme since you're updating it anyway (if it's not there already)

PR #394, opened 2024-04-09
ARIA Download NISAR GUNWs

bbuzzanga:
add download support for NISAR GUNWs

### DIFF
--- a/tools/bin/ariaDownload.py
+++ b/tools/bin/ariaDownload.py
@@ -317,7 +317,7 @@ class Downloader(object):
 
         ## buffer a bit for loose subsetting and speedup
         start = self.args.start + datetime.timedelta(days=-90)
-        end   = self.args.end + relativedelta(days=90)
+        end   = self.args.end + datetime.timedelta(days=90)
 
         if self.args.mission.upper() == 'S1':
             dct_kw = dict(dataset='ARIA S1 GUNW',
--- a/tools/bin/ariaDownload.py
+++ b/tools/bin/ariaDownload.py
@@ -15,6 +15,7 @@ import argparse
 import logging
 import shapely
 import warnings
+from getpass import getpass
 import asf_search
 
 import ARIAtools.util.log
@@ -31,16 +32,18 @@ def createParser():
     see: https://github.com/asfadmin/Discovery-asf_search
     """
     parser = argparse.ArgumentParser(
-        description='Command line interface to download GUNW products from '
-                    'the ASF DAAC. GUNW products are hosted at the NASA ASF '
-                     'DAAC.\nDownloading them requires a NASA Earthdata URS '
-                     'user login and requires users to add "GRFN Door (PROD)" '
-                     'and "ASF Datapool Products" to their URS approved '
-                     'applications.',
+        description='Command line interface to download Sentinel-1/NISAR GUNW '
+                    'products from the ASF DAAC. \nDownloading them requires a '
+                     'NASA Earthdata URS user login and requires users to add '
+                     '"GRFN Door (PROD)" and "ASF Datapool Products" to their '
+                     'URS approved applications. Access to NISAR products '
+                     'requires an Earthdata Bearer token from: '
+                     'https://urs.earthdata.nasa.gov/documentation/for_users/user_token',
         epilog='Examples of use:\n'
                 '\t ariaDownload.py --track 004 --output count\n'
                 '\t ariaDownload.py --bbox "36.75 37.225 -76.655 -75.928"\n'
-                '\t ariaDownload.py -t 004,077 --start 20190101 -o count',
+                '\t ariaDownload.py -t 004,077 --start 20190101 -o count'
+                '\t ariaDownload.py --mission NISAR -o count',
         formatter_class=argparse.RawDescriptionHelpFormatter)
 
     parser.add_argument(
@@ -76,6 +79,11 @@ def createParser():
         help='NASA Earthdata URS user password. Users must add "GRFN Door '
              '(PROD)" and "ASF Datapool Products" to their URS approved '
              'applications.')
+
+    parser.add_argument(
+        '--mission', default='S1', type=str.upper, choices=('S1', 'NISAR'),
+        help='Sentinel-1 (S1) or NISAR. Default is S1')
+
     parser.add_argument(
         '-l', '--daysless', dest='dayslt', default=math.inf, type=int,
         help='Take pairs with a temporal baseline -- days less than this value.')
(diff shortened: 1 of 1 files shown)


tools/bin/ariaDownload.py
```diff
@@ -302,15 +315,43 @@ def query_asf(self):
...
+                            session=session
+                        # processingLevel=asf_search.constants.GUNW_STD,
+                        # relativeOrbit=tracks,
+                        # flightDirection=flight_direction,
+                        # intersectsWith=bbox,
+                        # start=self.args.start,
+                        # end=self.args.end)
+                        )
+            scenes = asf_search.search(opts=search_opts, maxResults=250)
+            print (scenes[0])
```
alexfore: should this be logged instead of printed, other code in this block uses logger
`print (` should be `print(` (no whitespace before open parenthesis that starts the argument list of a function call)

tools/bin/ariaDownload.py
```diff
@@ -302,15 +315,43 @@ def query_asf(self):
...
+                        flightDirection=flight_direction,
+                        intersectsWith=bbox,
+                        start=st0,
+                        end=en0)
+            scenes = asf_search.geo_search(**dct_kw)
+
+        elif self.args.mission.upper() == 'NISAR':
+            session = asf_search.ASFSession()
+            session.auth_with_token(getpass('EDL Token:'))
+            LOGGER.info ('Token accepted.')
```
alexfore: no whitespace before open parenthesis that starts the argument list of a function call

tools/bin/ariaDownload.py
```diff
@@ -302,15 +315,43 @@ def query_asf(self):
...
-                      flightDirection=flight_direction,
-                      intersectsWith=bbox)
-        scenes = asf_search.geo_search(**dct_kw)
+
+        ## buffer a bit for loose subsetting and speedup
+        st0 = self.args.start + relativedelta(months=-3)
+        en0 = self.args.end + relativedelta(months=-3)
+
+        if self.args.mission.upper() == 'S1':
+            dct_kw = dict(dataset='ARIA S1 GUNW',
```
alexfore: misalignment

tools/bin/ariaDownload.py
```diff
@@ -15,6 +15,8 @@
 import logging
 import shapely
 import warnings
+from dateutil.relativedelta import relativedelta
```
alexfore: a comment on imports: In general, using the full path is more descriptive than taking parts of various packages as the namespace is truncated when using `from a import b`. IMO the full namespace is preferred due to the better descriptiveness but not always

tools/bin/ariaDownload.py
```diff
@@ -302,15 +315,43 @@ def query_asf(self):
...
 
-        dct_kw = dict(platform=asf_search.constants.SENTINEL1,
-                      processingLevel=asf_search.constants.GUNW_STD,
-                      relativeOrbit=tracks,
-                      flightDirection=flight_direction,
-                      intersectsWith=bbox)
-        scenes = asf_search.geo_search(**dct_kw)
+
+        ## buffer a bit for loose subsetting and speedup
+        st0 = self.args.start + relativedelta(months=-3)
```
alexfore: can we rename this vars `start` and `end` to match the other naming of attributes of `args` and arguments to `asf_search.geo_search`

tools/bin/ariaDownload.py
alexfore: Presumably the commented out lines will be uncommented later when other features are added? Can we add a TODO in the code here, is there some string we use in this repo to note features need to be added which is easy to search for

tools/bin/ariaDownload.py
alexfore: Is this a mistake here (why removing -3 months from both start and end)?

alexfore (changes requested):
I added some questions and a few style points

tools/bin/ariaDownload.py
bbuzzanga: agreed, but I think this one is commonly used

tools/bin/ariaDownload.py
bbuzzanga: not sure what it's supposed to be

alexfore: The llines are not aligned correctly

alexfore: dct_kw = dict(dataset='ARIA S1 GUNW',
                        processingLevel=asf_search.constants.GUNW_STD,


dataset and processingLevel should be aligned on subsequent lines

tools/bin/ariaDownload.py
alexfore: Recommend to use datetime.timedelta instead. We already use that in here. Also the pad should be a constant to avoid magic number anti pattern.

alexfore:
I've asked repeatedly for you to pay attention to the style of the code which you are committing. You have not done that so far. Please look at what you have done and ask yourself:
* Is what I have committed PEP8 compliant? 
* Is the use of whitespace consistent with the rest of the code in this module?
* Is what I have committed improving the software or adding more issues that need to be fixed in the future? (technical debt)

I think that style is extremely important. From my point of view imposing style requirements on this code base is the first step to making it a maintainable code base. If the main developers on this project do not agree I will not be able to help

It will be worth your time to memorize everything here: https://peps.python.org/pep-0008/

bbuzzanga:
I implemented all your requested changes. If you have more requests, point them out.

jhkennedy:
> I've asked repeatedly for you to pay attention to the style of the code which you are committing. You have not done that so far. Please look at what you have done and ask yourself:
> 
>     * Is what I have committed PEP8 compliant?
> 
>     * Is the use of whitespace consistent with the rest of the code in this module?
> 
>     * Is what I have committed improving the software or adding more issues that need to be fixed in the future? (technical debt)
> 
> 
> I think that style is extremely important. From my point of view imposing style requirements on this code base is the first step to making it a maintainable code base. If the main developers on this project do not agree I will not be able to help
> 
> It will be worth your time to memorize everything here: https://peps.python.org/pep-0008/


@alexfore, there seems to be a lot of unnecessary back and forth related to this that *should* be handled by CI/CD linting checks -- it's much better for everyone involved for a bot to be the arbiter. 

I suggest spending the effort on setting up a github action for flake8/pylint/ruff.

alexfore:
I would also love to integrate automated pep8 checks -- because I never want to have this conversation again. I do not like telling people that their code style is messy or inadequate, this clearly leads to conflict.

However, the fact remains that this PR has still contains pep8 violations -- whitespace and alignment.  My comments above still stand and I want to see them addressed before I will approve this PR. If someone is unwilling to improve I do not know what to say to that.  Maybe use pylint, pycodestyle, or flake8 and try to learn what the issues are. Me telling you or doing it myself is not going to help other people learn.

On my end I will figure out how to integrate the style checks I want in another PR.

dbekaert:
Hi @alexfore @jhkennedy I like the idea of a bot doing these sort of checks. I think they might even come with suggested fixes etc. Lets add that support in an upcoming PR and perhaps we can let the code-base run through it?

I would like to get the stable test suite in place, the dependency issue fixed for stability. This will allow to check against changes and see if something breaks. With that in place we could route the code through the both for suggested changes?

alexfore:
I am going to close this PR. We can open a new PR for this branch once I've integrated pep8 auto checks into this repo, and the author has addressed the issues raised by those checks.

alexfore:
This PR needs to be abandoned. These code changes need to start from the dev branch after I've merged dev_refactor.

PR #398, opened 2024-04-22
Dev refactor tests

alexfore:
PR for end to end regression tests:
* test ariaExtract
* test ariaDownload
* test ariaTSsetup


dbekaert (approved):
LGMT @alexfore. Thanks for adding this in. This will go a great way.

@sssangha lets track this nonfinite case and see if this can be resolved in the future.

PR #399, opened 2024-04-24
Dev refactor nisar

sssangha:
Push NISAR support changes

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -216,7 +216,7 @@ class MetadataQualityCheck:
                 # Unique bug-fix for bPerp layers with sign-flips
                 if ((self.prod_key == 'bPerpendicular' and
                     min(rsquaredarr) < 0.8 and max(std_errarr) > 0.1) and
-                    (negs_percent != 100 or negs_percent != 0)):
+                     (negs_percent != 100 or negs_percent != 0)):
 
                     # Circumvent Bperp sign-flip bug by comparing percentage of
                     # positive and negative values
@@ -945,7 +945,8 @@ def export_product_worker(
 
     # Extract/crop metadata layers
     if (any(':/science/grids/imagingGeometry' in s for s in product) or
-        any(':/science/LSAR/GUNW/metadata/radarGrid/' in s for s in product)):
+        any(':/science/LSAR/GUNW/metadata/radarGrid/' in s for s in product)
+         ):
         # make VRT pointing to metadata layers in standard product
         hgt_field, model_name, outname = prep_metadatalayers(
             outname, product, dem_expanded, layer, layers,
@@ -1421,7 +1422,7 @@ def finalize_metadata(outname, bbox_bounds, arrres, dem_bounds, prods_TOTbbox,
 
     metadatalyr_name = outname.split('/')[-2]
     if ((metadatalyr_name in GEOM_LYRS and version_check < '2_0_4')
-        and not is_nisar_file):
+         and not is_nisar_file):
         # create directory for quality control plots
         plots_subdir = os.path.abspath(os.path.join(outname, '../..',
                                        'metadatalyr_plots', metadatalyr_name))
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -326,9 +326,10 @@ def merged_productbbox(
     os.makedirs(workdir, exist_ok=True)
 
     # If specified, check if user's bounding box meets minimum threshold area
+    lyr_proj = int(metadata_dict[0]['projection'][0])
     if bbox_file is not None:
         user_bbox = ARIAtools.util.shp.open_shp(bbox_file)
-        overlap_area = ARIAtools.util.shp.shp_area(user_bbox)
+        overlap_area = ARIAtools.util.shp.shp_area(user_bbox, lyr_proj)
         if overlap_area < minimumOverlap:
             raise Exception(f'User bound box {bbox_file} has an area of only '
                             f'{overlap_area}km\u00b2, below specified '
@@ -343,13 +344,12 @@ def merged_productbbox(
         outname = os.path.join(workdir, pair_name + '.json')
 
         # Create union of productBoundingBox layers
-        for frame in scene["productBoundingBox"]:
-            prods_bbox = ARIAtools.util.shp.open_shp(
-                frame, 'productBoundingBox', 1)
+        for prods_bbox in scene["productBoundingBox"]:
             if os.path.exists(outname):
                 union_bbox = ARIAtools.util.shp.open_shp(outname)
                 prods_bbox = prods_bbox.union(union_bbox)
-            ARIAtools.util.shp.save_shp(outname, prods_bbox)
+            ARIAtools.util.shp.save_shp(
+                outname, prods_bbox, lyr_proj)
         scene["productBoundingBox"] = [outname]
 
     prods_TOTbbox = os.path.join(workdir, 'productBoundingBox.json')
@@ -364,17 +364,20 @@ def merged_productbbox(
     ind_max_area = sceneareas.index(max(sceneareas))
     product_bbox = product_dict[ind_max_area]['productBoundingBox'][0]
     ARIAtools.util.shp.save_shp(
-        prods_TOTbbox_metadatalyr, ARIAtools.util.shp.open_shp(product_bbox))
+        prods_TOTbbox_metadatalyr, ARIAtools.util.shp.open_shp(product_bbox),
+        lyr_proj)
 
     # Initiate intersection file with bbox, if bbox specified
     if bbox_file is not None:
         ARIAtools.util.shp.save_shp(
-            prods_TOTbbox, ARIAtools.util.shp.open_shp(bbox_file))
+            prods_TOTbbox, ARIAtools.util.shp.open_shp(bbox_file),
+            lyr_proj)
 
     # Initiate intersection with largest scene, if bbox NOT specified
     else:
         ARIAtools.util.shp.save_shp(
-            prods_TOTbbox, ARIAtools.util.shp.open_shp(product_bbox))
+            prods_TOTbbox, ARIAtools.util.shp.open_shp(product_bbox),
+            lyr_proj)
 
     rejected_scenes = []
     for scene in product_dict:
@@ -390,9 +393,11 @@ def merged_productbbox(
             total_bbox_metadatalyr = total_bbox_metadatalyr.union(prods_bbox)
 
             # Save to file
-            ARIAtools.util.shp.save_shp(prods_TOTbbox, total_bbox)
             ARIAtools.util.shp.save_shp(
-                prods_TOTbbox_metadatalyr, total_bbox_metadatalyr)
+                prods_TOTbbox, total_bbox, lyr_proj)
+            ARIAtools.util.shp.save_shp(
+                prods_TOTbbox_metadatalyr, total_bbox_metadatalyr,
+                lyr_proj)
 
         # Generate footprint for the common intersection of all products
         else:
@@ -414,7 +419,8 @@ def merged_productbbox(
                 os.remove(scene_obj)
 
             else:
-                overlap_area = ARIAtools.util.shp.shp_area(prods_bbox)
+                overlap_area = ARIAtools.util.shp.shp_area(
+                    prods_bbox, lyr_proj)
 
                 # Kick out scenes below specified overlap threshold
                 if overlap_area < minimumOverlap:
@@ -424,7 +430,8 @@ def merged_productbbox(
                     os.remove(scene_obj)
 
                 else:
-                    ARIAtools.util.shp.save_shp(prods_TOTbbox, prods_bbox)
+                    ARIAtools.util.shp.save_shp(
+                        prods_TOTbbox, prods_bbox, lyr_proj)
 
                     # Need to track bounds of max extent
                     # to avoid metadata interpolation issues
@@ -432,7 +439,8 @@ def merged_productbbox(
                         ARIAtools.util.shp.open_shp(
                             scene['productBoundingBox'][0]))
                     ARIAtools.util.shp.save_shp(
-                        prods_TOTbbox_metadatalyr, total_bbox_metadatalyr)
+                        prods_TOTbbox_metadatalyr, total_bbox_metadatalyr,
+                        lyr_proj)
 
     # Remove scenes with insufficient overlap w.r.t. bbox
     if rejected_scenes != []:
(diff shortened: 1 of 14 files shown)


tools/ARIAtools/extractProduct.py
```diff
@@ -209,12 +208,14 @@ def __run__(self):
...
-                    self.data_array_band == no_data_value, self.data_array_band)
+                    self.data_array_band == no_data_value,
+                    self.data_array_band)
                 negs_percent = ((self.data_array_band < 0).sum()
                                 / self.data_array_band.size) * 100
 
                 # Unique bug-fix for bPerp layers with sign-flips
-                if (self.prod_key == 'bPerpendicular' and min(rsquaredarr) < 0.8
+                if (self.prod_key == 'bPerpendicular'
+                        and min(rsquaredarr) < 0.8
```
alexfore: prefer and on prev line, and min should have same column as self

tools/ARIAtools/extractProduct.py
```diff
@@ -691,16 +741,28 @@ def handle_epoch_layers(
         ifg = product_dict[1][i[0]][0]
         outname = os.path.abspath(os.path.join(workdir, ifg))
 
+        # capture model if tropo product
+        model_name = ''
```
alexfore: Prefer using None over empty string, unless it needs to be a string type -- None is more pythonic and will be more clear to any future readers of the code. Then later the check should be is not None vs != ''

tools/ARIAtools/extractProduct.py
```diff
@@ -860,15 +932,18 @@ def export_product_worker(
     outname = os.path.abspath(os.path.join(workdir, ifg_tag))
 
     # Extract/crop metadata layers
-    if any(":/science/grids/imagingGeometry" in s for s in product):
+    if any(':/science/grids/imagingGeometry' in s for s in product) \
```
alexfore: Prefer to use parens instead of backslash for line continuation (according to pep8 not just me)

tools/ARIAtools/extractProduct.py
```diff
@@ -1019,14 +1077,33 @@ def export_products(
...
+        'layers': layers, 'prods_TOTbbox': prods_TOTbbox,
+        'proj': int(epsg_code), 'dem': dem_expanded, 'lat': lat, 'lon': lon,
+        'mask': mask, 'verbose': verbose, 'multilooking': multilooking,
+        'rankedResampling': rankedResampling, 'num_threads': num_threads}
+
+    # track if product stack is NISAR GUNW or not
+    nisar_file = False
+    range_correction = True
+    track_fileext = full_product_dict[0]['unwrappedPhase'][0]
+    if len(track_fileext.split('.h5')) > 1:
```
alexfore: A different check will be more obvious to any reader of the code (i.e. .endswith() versus .split), it more clear == less cognitive load

tools/ARIAtools/extractProduct.py
```diff
@@ -1141,7 +1242,9 @@ def export_products(
 
     # If specified, extract ionosphere long wavelength
     ext_corr_lyrs += ['ionosphere']
-    if 'ionosphere' in layers:
+    user_lyrs = list(
```
alexfore: Why can't we use `if 'ionosphere' in layers:` here? Unless I'm mistaken, this yields exactly the same conditional as if `'ionosphere' in layers` but the cognitive load of this `set.intersection(*map(set, [layers, ['ionosphere']]))` expression is very high

tools/ARIAtools/extractProduct.py
```diff
@@ -1297,13 +1403,19 @@ def finalize_metadata(outname, bbox_bounds, arrres, dem_bounds, prods_TOTbbox, d
...
+            basename = os.path.basename(i.split('"')[1])
+            v_num = basename.split('_')[-1][:-3]
+            v_num = '.'.join(v_num)
         version_check.append(v_num)
     version_check = min(version_check)
 
     metadatalyr_name = outname.split('/')[-2]
-    if metadatalyr_name in GEOM_LYRS and version_check < '2_0_4':
+    if (metadatalyr_name in GEOM_LYRS and version_check < '2_0_4') \
+            and not nisar_file:
```
alexfore: prefer parens over backslash (pep8), the continued line should align with metadatalyr_name

tools/ARIAtools/extractProduct.py
```diff
@@ -1677,23 +1791,28 @@ def gacos_correction(full_product_dict, gacos_products, bbox_file,
                 # applicable)
                 elif '-' in tropo_rsc_dict['TIME_OF_DAY'][0]:
                     tropo_rsc_dict['TIME_OF_DAY'] = [
-                        datetime.datetime.strptime(m[:10] + '-' +
-                        m[11:].split('.')[0] + '-'
-                        + str(round(float('0.' + m[11:].split('.')[1]) * 60)),
-                        "%Y-%m-%d-%H-%M")
-                        for m in tropo_rsc_dict['TIME_OF_DAY']
-                    ]
+                        datetime.datetime.strptime(
```
alexfore, line 1804: These expressions for `tropo_rsc_dict['TIME_OF_DAY']` are all too complicated and should be refactored so less expressions are being done inline

tools/ARIAtools/product.py
```diff
@@ -12,11 +12,15 @@
...
 import itertools
+import h5py
 import netCDF4
 import numpy as np
 import osgeo
+from pyproj import Transformer
 
 import shapely.geometry
+from shapely.ops import transform
+from shapely.wkt import loads
```
alexfore: Please use the full namespace and avoid the from X import Y pattern. A reader of the code will not know where these objects are coming from (Transformer, transform, loads)

tools/bin/ariaTSsetup.py
```diff
@@ -245,7 +242,7 @@ def generate_stack(aria_prod, stack_layer, output_file_name,
 
     # handle individual epochs if external correction layer
     if (domain_name in ARIA_EXTERNAL_CORRECTIONS or
-        domain_name in ARIA_TROPO_MODELS):
+            domain_name in ARIA_TROPO_MODELS):
```
alexfore: domain_name should be aligned vertically

tools/ARIAtools/extractProduct.py
```diff
@@ -1659,16 +1772,17 @@ def gacos_correction(full_product_dict, gacos_products, bbox_file,
...
+                    datetime.datetime.strptime(
+                        os.path.basename(j)[: 4] + '-' + os.path.
+                        basename(j)[4: 6] + '-' + os.path.basename(j)[6: 8] +
+                        '-' + m[11:],
+                        "%Y-%m-%d-%H:%M:%S.%f") for m in metadata_dict[0][0]]
                 # Get tropo product UTC times
                 tropo_rsc_dict = {}
-                tropo_rsc_dict['TIME_OF_DAY'] = open(j[:-4] + '.rsc',
-                                                     'r').readlines()[-1].split()[1].split('UTC')[:-1]
+                tropo_rsc_dict['TIME_OF_DAY'] = open(
```
alexfore, line 1793: Same comment as below (cognitive load of this expression is too high)

tools/ARIAtools/extractProduct.py
```diff
@@ -1677,23 +1791,28 @@ def gacos_correction(full_product_dict, gacos_products, bbox_file,
...
-                    ]
+                        datetime.datetime.strptime(
+                            m[: 10] + '-' + m[11:].split('.')[0] + '-' +
+                            str(
+                                round(
+                                    float('0.' + m[11:].split('.')[1]) * 60)),
+                            "%Y-%m-%d-%H-%M")
+                        for m in tropo_rsc_dict['TIME_OF_DAY']]
                 else:
                     tropo_rsc_dict['TIME_OF_DAY'] = [
```
alexfore, line 1812: Same comment as above (cognitive load of this expression is too high)

tools/ARIAtools/product.py
```diff
@@ -297,7 +305,7 @@ def __init__(self, filearg, bbox=None, workdir='./', num_threads=1,
         tmp_files = self.files.copy()
         for f in tmp_files:
             ext = os.path.splitext(f)[1].lower()
-            if not ext == '.nc':
+            if not ext == '.nc' and not ext == '.h5':
```
alexfore: ext not in ['.nc', '.h5'] is a more sustainable pattern here

tools/ARIAtools/product.py
```diff
@@ -321,7 +329,9 @@ def __init__(self, filearg, bbox=None, workdir='./', num_threads=1,
                     'products: Linux kernel >=4.3 and libnetcdf >=4.5')
 
         # check if local file reader is being captured as netcdf
-        if any("https://" not in i for i in self.files):
+        check_for_urls = any("https://" not in i for i in self.files)
```
alexfore, line 332: What is being determined by these lines of code? If this is a nisar file? If the file ends with .h5? If the file starts with https://?

IMO this can be made much more clear / lower cognitive load

tools/ARIAtools/product.py
```diff
@@ -387,24 +425,36 @@ def __readproduct__(self, fname):
...
         # Get standard product version from file
         # version accessed differently between URL vs downloaded product
-        version = str(
-            osgeo.gdal.Open(fname).GetMetadataItem('NC_GLOBAL#version'))
-        if version == 'None':
-            LOGGER.warning(
-                '%s is not a supported file type... skipping', fname)
-            return []
+        # vs NISAR and S1 GUNWs
+        if basename.split('_')[0] == 'NISAR':
```
alexfore: This could be basename.startswith('NISAR_')  which is more simple/clear/lower cognitive load

tools/ARIAtools/product.py
```diff
@@ -584,6 +648,7 @@ def __mappingVersion__(self, fname, version):
                     'centerLatitude'][:-1])
 
             # hardcoded keys for a given sensor
+            rdrmetadata_dict['projection'] = self.projection
             if basename.split('-')[0] == 'S1':
```
alexfore: .startswith('S1') is more explicit than split('-')[0] == 'S1'

alexfore (changes requested):
Thanks @sssangha, I think this looks pretty good. I do have a lot of comments, please don't feel that I am being overly critical. Consistent style (PEP8) and reduction of cognitive load (replace complicated expressions with simple ones) are key to maintainability IMO.

Some general ideas:
* `nisar_file`: This seems to be a bool, where it is True if it is a nisar file and False otherwise. Recommend to rename this var `is_nisar_file` or similar so that the name aligns with the meaning (is_nisar_file is more clearly a True/False value).
* It seems like line breaks were added to some lines that were not needed to keep line length under 79
* Recommend that we only use one of `ds = None` or `del ds` for the attempts to force GDAL to close open files?. I see both being used and that is confusing. Prefer ds = None (generally should avoid using del as it does not actually garbage collect RAM at that point).

All of my other comments are either pep8 or aimed at reducing cognitive load by making expressions more simple and easy to understand.

tools/ARIAtools/extractProduct.py
sssangha: This was one of the items I cleared up in the `dev_refactor_mskupdate` branch.. So I'll leave this as is here

tools/ARIAtools/extractProduct.py
sssangha: This whole `gacos_correction` function will go away completely in a future PR. I want to remove that, plus other vestigial functions/scripts then.

tools/ARIAtools/product.py
sssangha: @alexfore I tried this, but get this error:
```
>>> import shapely.ops.transform
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'shapely.ops.transform'; 'shapely.ops' is not a package
```

How about this instead so users can get an idea of the provenance as you advise:
from shapely.ops import transform as shapely_ops_transform

alexfore: I'm saying do this
```
import shapely.geometry
import shapely.ops
import shapely.wkt
```
and use the full namespace in the function calls (`transform` and `loads`), also may not need `import shapely.geometry` as `import shapely` seems to import `shapely.geometry` already

sssangha: Ok done. I specified as `import shapely.geometry` instead of `import shapely`

sssangha:
@alexfore, I worked to satisfy some suggestions from the pep8 bot, namely indentation of `if` statements spanning multiple lines.

From your commit history, in these cases it looks like you had indented the lines by one space. However, the bot kept raising a flag for such cases, which only went away when I increased the indent to 4 spaces as reflected [here](https://www.flake8rules.com/rules/E129.html). I would like your advice on this, as I want to be consistent with what we do and avoid overriding any changes unnecessarily.

alexfore:
The bot seems to insist on the last of the three options which pep8 says is acceptable here: https://peps.python.org/pep-0008/#indentation, (cut from that page, pep8 says all of these are acceptable):
```
# No extra indentation.
if (this_is_one_thing and
    that_is_another_thing):
    do_something()

# Add a comment, which will provide some distinction in editors
# supporting syntax highlighting.
if (this_is_one_thing and
    that_is_another_thing):
    # Since both conditions are true, we can frobnicate.
    do_something()

# Add some extra indentation on the conditional continuation line.
if (this_is_one_thing
        and that_is_another_thing):
    do_something()
```
I suppose I've never used or realized the last one was acceptable, I always used the second or first, depending on if the code needed a comment. There is quite a bit of discussion about this in various places online -- E125 and E129 both violate pep8, I could add a config file for pep8speaks which will disable E125 and E129, what do you think?

sssangha:
> The bot seems to insist on the last of the three options which pep8 says is acceptable here: https://peps.python.org/pep-0008/#indentation, (cut from that page, pep8 says all of these are acceptable):
> 
> ```
> # No extra indentation.
> if (this_is_one_thing and
>     that_is_another_thing):
>     do_something()
> 
> # Add a comment, which will provide some distinction in editors
> # supporting syntax highlighting.
> if (this_is_one_thing and
>     that_is_another_thing):
>     # Since both conditions are true, we can frobnicate.
>     do_something()
> 
> # Add some extra indentation on the conditional continuation line.
> if (this_is_one_thing
>         and that_is_another_thing):
>     do_something()
> ```
> 
> I suppose I've never used or realized the last one was acceptable, I always used the second or first, depending on if the code needed a comment. There is quite a bit of discussion about this in various places online -- E125 and E129 both violate pep8, I could add a config file for pep8speaks which will disable E125 and E129, what do you think?

@alexfore ok understood, thanks for the clarification!

And yeah sure, if you could add a config file to disable E125 and E129, that would be great!

alexfore (approved):

PR #401, opened 2024-05-01
Dev refactor

alexfore:
Partial refactor of ARIA-tools
* More domain specific knowledge needed to finish refactoring
* add util submodule
* fix executable permissions
* Rename methods in util.shp
* Add default args to util.shp.open / save methods
* Clean up logging and add --log-level command line args
* improve clarity of main computational loop
* Remove lat/lon from export_products_worker interface
* Add ability to run export_products in parallel with GNU parallel
* Add multiproc_method argument to export_products with three options:
  * gnu_parallel: will use subjobs, executed in parallel, to process export_products_worker
  * threads: will use multiple threads, faster startup/tear down but has the python GIL issue
  * single: will run single-threaded


alexfore:
I'm not going to address the style issues in ariaPlot at this time -- they are mostly related to the nested list comps which I lack the energy to decompose.

PR #404, opened 2024-05-13
Input shapefile not correctly parsed

sssangha:
If a valid shapefile path was passed, only a `poly` variable would be returned. No `W,S,N,E` variables would be initiated in the if statement and the function would crash


alexfore (approved):
Thanks

PR #408, opened 2024-05-14
Remove relax from build

sssangha:
Relax 4 is legacy code that has been removed and this PR is the works to remove it in the requirements, the build, as well as reference to it in the ARIA-tools code base.


alexfore:
What is the goal of this PR? (this should be in the description)

When I grep for relax in ARIA-tools I find more hits than what was touched in this commit:
```
╭─fore@frode  /u/frode-z/fore/aria/ARIA-tools git:‹dev› 
╰─> grep -r relax .                                                                  [135/11:41:26]
./setup.py:# Path where relax third party is download to
./setup.py:relaxPath = 'tools/thirdParty/Min-Cost-Flow-Class'
./setup.py:if os.path.isdir(relaxPath):
./setup.py:            'tools/bindings/relaxIVdriver.cpp',
./setup.py:            os.path.join(relaxPath, 'RelaxIV/RelaxIV.C')],
./setup.py:            'tools/include', os.path.join(relaxPath, 'MCFClass'),
./setup.py:            os.path.join(relaxPath, 'OPTUtils'),
./setup.py:            os.path.join(relaxPath, 'RelaxIV')])
./tools/bindings/unwcompmodule.cpp:#include "relaxIVdriver.h"
./tools/bindings/unwcompmodule.cpp:PyObject* relaxIVwrapper_C(PyObject* self, PyObject* args)
./tools/bindings/relaxIVdriver.cpp:#include <relaxIVdriver.h>
./tools/include/unwcompmodule.h:    PyObject *relaxIVwrapper_C(PyObject *self,PyObject *args);
./tools/include/unwcompmodule.h:    {"relaxIVwrapper_Py",relaxIVwrapper_C,METH_VARARGS," "},
./tools/include/relaxIVdriver.h:#ifndef relaxivdriver_h 
./tools/include/relaxIVdriver.h:#define relaxivdriver_h
./tools/ARIAtools/unwrapStitching.py:        # and relaxIV MCF solver is used by default
./tools/ARIAtools/phaseMinimization.py:    #  Phase unwrap using relax IV
./tools/ARIAtools/phaseMinimization.py:        return unwcomp.relaxIVwrapper_Py(fileName)
```

sssangha:
We wanted to get rid of `relax` from the build because it is only leveraged by the vestigial unwrapping functionality. So those files you pointed to should all be removed altogether.

So then imports like [this](https://github.com/aria-tools/ARIA-tools/blob/aria-setup_removerelax/tools/ARIAtools/extractProduct.py#L33) would also have to go away.

Would such changes be appropriate for this PR @alexfore ?

dbekaert:
@sssangha, your PR should touch upon more than just the set-up. Also remove reference to relax at other locations. Relax is fully absolute now.

Note that means like Relax compiler code etc, as well as updating the acutal code where there are made comments to relax, remove relax from any input options etc.

sssangha:
Ok @dbekaert went ahead and pushed the appropriate updates.

@alexfore I updated the description and removed the vestigial references to the obsolute `relax` module and associated stitching routine.

sssangha:
@alexfore and @dbekaert ok, so to confirm I no longer get references to the vestigial `relax` module and the `unwrapStitching` scripts. It looks good to go on my end.
```
(base) ➜  relax_remove grep -r relax ARIA-tools          
(base) ➜  relax_remove 
(base) ➜  relax_remove grep -r unwrapStitching ARIA-tools
(base) ➜  relax_remove 
```

dbekaert:
@sssangha we should check if this all compiles well and if the test classes still all run fine. 
@alexfore dont think the test suite is being triggered here. Might be related to forks and different rule?

alexfore:
I do not have the tests setup as a github action, I'm not quite sure how to set that up. It is run with `pytest` in `test/regression`. I also disabled (it is still there and can be enabled again) the existing test since it always failed

alexfore:
This passes all of the tests in `tests/regression`

sssangha:
Ok great to hear! @alexfore , please advise whether any further action is needed for this PR. Thanks

Issue #420, opened 2024-06-12
[BUG] TSSetup bug when extracting ionosphere layer


sssangha:
**Describe the bug**
When specifying the extraction of the `ionosphere` layer through `ariaTSSetup.py`, an error is raised from the part of the code which checks for consistency in extracted layer dimensions.

**To Reproduce**
1. Download:
`ariaDownload.py --num_threads 16 -b '33.4 34.5 -119.0 -117.4'  -s 20190117  -e 20190210 -t 64 --version 3_0_1 -o Download `

2. Command used:
`ariaTSsetup.py -f "products/*.nc" -b '33.6 34.3 -118.5 -117.5' --verbose -w test_iono -l ionosphere --log-level debug`

3. Error Output:
```
Traceback (most recent call last):
  File "/u/trappist-r0/ssangha/conda_installation/miniforge/miniforge/ARIA-tools_dev/tools/bin/ariaTSsetup.py", line 634, in <module>
    main()
  File "/u/trappist-r0/ssangha/conda_installation/miniforge/miniforge/ARIA-tools_dev/tools/bin/ariaTSsetup.py", line 570, in main
    ARIAtools.util.vrt.dim_check(ref_arr_record, prod_arr_record)
  File "/u/trappist-r0/ssangha/conda_installation/miniforge/miniforge/ARIA-tools_dev/tools/ARIAtools/util/vrt.py", line 542, in dim_check
    raise Exception(
Exception: Inconsistent product dims between products test_iono/unwrappedPhase/20190123_20190117 and None: check respective width (1201, None) and height (841, None) and geotrans ((-118.500785933, 0.000833333000000001, -0.0, 34.300819613, -0.0, -0.000833333000000001), None)
```

**Desktop (please complete the following information):**
 - ARIA-tools git tag: `commit 267c84436dd902df2266a9581543f676eb065cb7 (HEAD -> dev, tag: v1.1.8, origin/dev, origin/HEAD)`

**Proposed fix**
I have a working patch, just raising this issue ticket to record and reference it.

PR #421, opened 2024-06-12
Fix bug when extracting iono and tropo layer

sssangha:
Addressing issue #420 , where an error is raised when just either the ionosphere and tropo layer are extracted because the dimensions were not properly passed and accounted for.

Specifically, the program checks for internal consistency between the dimensions of all output layers. This information was not initialized and passed for the ionosphere and tropo layers, so when this check is executed the program fails.

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -1195,7 +1195,10 @@ def export_products(
 
             # track valid files
             prev_outname = os.path.abspath(
-                os.path.join(workdir, i, product_dict[1][0][0]))
+                os.path.join(workdir,
+                             i.split('_')[-1],
+                             product_dict[1][0][0])
+            )
             if os.path.exists(prev_outname + '.vrt'):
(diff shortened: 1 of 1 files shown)


tools/ARIAtools/extractProduct.py
```diff
@@ -1195,7 +1195,10 @@ def export_products(
 
             # track valid files
             prev_outname = os.path.abspath(
-                os.path.join(workdir, i, product_dict[1][0][0]))
+                os.path.join(workdir,
+                             i.split('_')[-1],
```
alexfore: Can we move `i.split('_')[-1]` out of the `os.path.join` expression and into its own expression (i.e. `tag = i.lstrip('_')` then include `tag` in `os.path.join`? Then include a comment about what this is doing (removing leading underscore)

alexfore (approved):

tools/ARIAtools/extractProduct.py
sssangha: Done!

Issue #422, opened 2024-06-18
[BUG] bug when passing `NUM_THREADS` options through gdalWarp


sssangha:
**Describe the bug**
With the latest increment in gdal (`3.9.0`), the `NUM_THREADS` variable now raises an error when passed through gdalWarp.

**To Reproduce**
1. Download:
`ariaDownload.py --num_threads 16 -b '26.34 27.86 36.767 39.874' -s 20150401 -e 20150501 -t 14 -o Download `

2. Command used:
`ariaTSsetup.py -f "products/*.nc" -b '26.34 27.86 36.767 39.874' --verbose -w test_dir_all_redo -l all --log-level debug`

3. Error Output:
```
(ARIA-tools_dev) [ssangha@trappist issue_420_ariatssetup]$ ariaTSsetup.py -f "products/*.nc" -b '33.6 34.3 -118.5 -117.5' --verbose -w test_dir_extract -l amplitude --log-level debug
2024-06-17 14:43:44,679 - ariaTSsetup.py - INFO - ARIAtools version: 1.1.8
2024-06-17 14:43:44,679 - ariaTSsetup.py - INFO - Time-series Preparation Function
2024-06-17 14:43:44,679 - ariaTSsetup.py - INFO - Thread count specified for gdal multiprocessing = 2
2024-06-17 14:43:44,679 - ariaTSsetup.py - INFO - Building ARIA product instance
2024-06-17 14:43:44,781 - ARIAtools.product - INFO - Shapefile test_dir_extract/user_bbox.json created for input user bounds
2024-06-17 14:43:45,100 - ARIAtools.product - INFO - All (8) GUNW products meet spatial bbox criteria.
2024-06-17 14:43:45,101 - ARIAtools.product - INFO - Group GUNW products into spatiotemporally continuous interferograms.
2024-06-17 14:43:45,103 - ARIAtools.product - INFO - All (3) interferograms are spatially continuous.
2024-06-17 14:43:45,103 - ariaTSsetup.py - INFO - Extracting and merging product bounding boxes
gdal_warp_kwargs {'format': 'MEM', 'multithread': True, 'dstSRS': 'EPSG:4326', 'options': ['NUM_THREADS=2']}
Traceback (most recent call last):
  File "/u/trappist-r0/ssangha/conda_installation/miniforge/miniforge/ARIA-tools_dev/tools/bin/ariaTSsetup.py", line 632, in <module>
    main()
[...]

PR #424, opened 2024-06-18
Add patch to handle gdal update to num_threads option

sssangha:
Addresses issue #422

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -11,7 +11,7 @@ dependencies:
   - asf_search
   - dask
   - dem_stitcher>=2.5.0
-  - gdal>=3.4.1
+  - gdal>=3.7.0
   - h5py
   - joblib
   - matplotlib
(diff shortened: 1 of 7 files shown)


alexfore:
I verified this passes the regression tests.

My only comment would be regarding if these code changes are compatible with older versions of GDAL / should we modify the version specified in `environment.yml` / `requirements.txt`?

alexfore:
config_options was added in 3.7.0 per this: https://github.com/OSGeo/gdal/blob/v3.7.0/NEWS.md

sssangha:
> config_options was added in 3.7.0 per this: https://github.com/OSGeo/gdal/blob/v3.7.0/NEWS.md

You're right @alexfore , we should update the requirements to enforce compatibility. I've done this now.

alexfore (approved):

PR #425, opened 2024-07-06
fix seaborn issue in plotting of shapefiles

bbuzzanga:


alexfore:
The changing of the function name seems to be a mistake: (`def shp_area(...` was changed to `def shp_are(...`)

alexfore:
Per the below link, seaborn-v0_8 was introduced in matplotlib 3.6.0, what version is raising this issue? https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.6.0.html

Another possible resolution is to bump the required version of matplotlib

bbuzzanga:
the problem is with `with plt.style.context(('seaborn')):` which was deprecated at some point. The change I put in should be backward compatible.

alexfore (approved):

PR #427, opened 2024-07-19
Check DEM provided when extracting layers for which necessary

rzinke:
Added function to check that a valid DEM is specified or downloaded when requesting ariaExtract.py to extract 'all' or any of 'bPerpendicular', 'bParallel', 'incidenceAngle', 'lookAngle', 'azimuthAngle', or 'solidEarthTide', for which a DEM is necessary.

### DIFF
--- a/tools/bin/ariaExtract.py
+++ b/tools/bin/ariaExtract.py
@@ -153,22 +153,21 @@ def main():
                   else layer for layer in layers]
 
         # list of layers requiring DEM for extraction
-        layers_requiring_dem = ['all',
+        LAYERS_REQUIRING_DEM = {'all',
                                 'bPerpendicular',
                                 'bParallel',
                                 'incidenceAngle',
                                 'lookAngle',
                                 'azimuthAngle',
-                                'solidEarthTide']
+                                'solidEarthTide'}
 
         # check that DEM is specified depending on layers requested
-        intersecting_layers = [layer for layer in layers
-                               if layer in layers_requiring_dem]
-        if intersecting_layers != []:
-            assert args.demfile is not None, (
-                'A valid DEM must be specified when extracting any of '
-                f"{' '.join(intersecting_layers)}"
-            )
+        if len(LAYERS_REQUIRING_DEM.intersection(layers)) > 0:
+            if args.demfile is None:
+                error_msg = 'A valid DEM must be specified when extracting ' \
+                            'any of %s' % ', '.join(LAYERS_REQUIRING_DEM)
+                LOGGER.error(error_msg)
+                raise Exception(error_msg)
 
     # if user bbox was specified, file(s) not meeting imposed spatial criteria
     # are rejected.
--- a/tools/bin/ariaExtract.py
+++ b/tools/bin/ariaExtract.py
@@ -142,6 +142,31 @@ def main():
     logging.basicConfig(level=log_level, format=ARIAtools.util.log.FORMAT)
     LOGGER.info('Extract Product Function')
 
+    # Check whether all necessary inputs were specified.
+    # some products require a DEM to extract -- if any of those are requested,
+    # ensure that a valid DEM is specified
+    if args.layers:
+        # format list of layers
+        layers = list(args.layers.split(','))
+        layers = [i.replace(' ', '') for i in layers]
+        layers = [layer.replace(layer, 'all')
+                  for layer in layers if layer.lower() == 'all']
+
+        # list of layers requiring DEM for extraction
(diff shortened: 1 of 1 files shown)


sssangha:
Thanks @rzinke!

I issued a patch to capture a bug with the way the `all` keyword was handled. In the initial commit, you had dummy-proofed it by forcing it to be lower-case. But the way it was implemented kicked out all keywords that weren't `all`. I.e. if I specified just SET then an empty list would be passed.

Also, I made some minor tweaks to meet pep8 suggestions. @alexfore , could you please advise on the implementation on this end? Thanks

Here are some simple tests to test this all out:
```
# get products
wget https://gunw-development-testing.s3.us-west-2.amazonaws.com/unavco_2024/S1-GUNW-A-R-124-tops-20180502_20180408-043026-00157W_00018N-PP-9909-v3_0_1.nc .
wget https://gunw-development-testing.s3.us-west-2.amazonaws.com/unavco_2024/S1-GUNW-A-R-124-tops-20180502_20180408-043052-00157W_00019N-PP-7296-v3_0_1.nc .

# test SET extraction without specifying a valid dem
ariaExtract.py -f "products/*.nc" -l solidEarthTide -w test_set_nodem

# test SET extraction with DEM download specified
ariaExtract.py -f "products/*.nc" -l solidEarthTide -w test_set_dem -d Download
```

tools/bin/ariaExtract.py
```diff
@@ -142,6 +142,34 @@ def main():
     logging.basicConfig(level=log_level, format=ARIAtools.util.log.FORMAT)
     LOGGER.info('Extract Product Function')
 
+    # Check whether all necessary inputs were specified.
+    # some products require a DEM to extract -- if any of those are requested,
+    # ensure that a valid DEM is specified
+    if args.layers:
```
alexfore: Recommend to use the more pythonic version of this: `args.layers is not None:`

tools/bin/ariaExtract.py
```diff
@@ -142,6 +142,34 @@ def main():
...
+        # format list of layers
+        layers = list(args.layers.split(','))
+        layers = [i.replace(' ', '') for i in layers]
+        layers = ['all' if layer.lower() == 'all'
+                  else layer for layer in layers]
+
+        # list of layers requiring DEM for extraction
+        layers_requiring_dem = ['all',
```
alexfore: Recommend `LAYERS_REQUIRING_DEM` here for constant (constants in UPPER_CASE)

tools/bin/ariaExtract.py
```diff
@@ -142,6 +142,34 @@ def main():
...
+                                'bPerpendicular',
+                                'bParallel',
+                                'incidenceAngle',
+                                'lookAngle',
+                                'azimuthAngle',
+                                'solidEarthTide']
+
+        # check that DEM is specified depending on layers requested
+        intersecting_layers = [layer for layer in layers
```
alexfore: IMO let's use sets and intersection here (more expressive):
`if len(set(LAYERS_REQUIRING_DEM).intersection(layers)) == 0:`
or declare LAYERS_REQUIRING_DEM as a set (use curly braces), then it will be 
`if len(LAYERS_REQUIRING_DEM.intersection(layers)) == 0:`

tools/bin/ariaExtract.py
```diff
@@ -142,6 +142,34 @@ def main():
...
+                               if layer in layers_requiring_dem]
+        if intersecting_layers != []:
+            assert args.demfile is not None, (
```
alexfore: Should this be an assertion, or should we logically and the assertion with the conditional on the line above then raise an Exception and issue a `LOGGER.error` message here if passes conditional? IMO this would be more simple to understand but same behavior

alexfore (changes requested):
I have some quick comments above, should be minor changes.

rzinke (commented):
Thank you both. I will implement the suggested changes, except that the ionosphere layer doesn't need a DEM (I checked last week when modifying the notebook).

rzinke:
Updates made [0b94a0d](0b94a0dadbbd2e1d0b12dba4e970ce5ec595c564)
@alexfore could you please review?

tools/bin/ariaExtract.py
alexfore: @rzinke Just this last comment here, thanks!

sssangha: Done, thanks @alexfore !

alexfore (approved):
Thanks!

PR #431, opened 2024-07-24
Push changes to dummy-proof tropo options

sssangha:
Access test products for this case example as so:
```
wget https://gunw-development-testing.s3.us-west-2.amazonaws.com/unavco_2024/S1-GUNW-A-R-124-tops-20180502_20180408-043026-00157W_00018N-PP-9909-v3_0_1.nc .
wget https://gunw-development-testing.s3.us-west-2.amazonaws.com/unavco_2024/S1-GUNW-A-R-124-tops-20180502_20180408-043052-00157W_00019N-PP-7296-v3_0_1.nc .
```

If a user specifies the extraction of a troposphere layer, but neglects to specify an argument for `--tropo_models`, error out early instead of silently failing to extract it.

To test this, run the following command:
```
ariaExtract.py -f "*.nc" -w tropo_test1 -d Download -l troposphereTotal
```

If a user specifies a tropo model, but neglects to specify the extraction of a valid troposphere layer with the '-l' option, error out early instead of silently failing to extract it.

To test this, run the following command:
```
ariaExtract.py -f "*.nc" -w tropo_test2 -d Download --tropo_models ERA5
```

Finally, if a user specifies a tropo model that doesn't exist, catch this early and error out

To test this, run the following command:
```
ariaExtract.py -f "*.nc" -w tropo_test3 -d Download --tropo_models FAKE_MODEL -l troposphereTotal
```

### DIFF
--- a/tools/ARIAtools/product.py
+++ b/tools/ARIAtools/product.py
@@ -370,7 +370,7 @@ class Product:
                             'as it is not one of the '
                             'following valid models: %s' % (
                                 i, ', '.join(
-                                ARIAtools.constants.ARIA_TROPO_INTERNAL)
+                                    ARIAtools.constants.ARIA_TROPO_INTERNAL)
                             )
                         )
                         LOGGER.error(error_msg)
--- a/tools/ARIAtools/product.py
+++ b/tools/ARIAtools/product.py
@@ -9,7 +9,6 @@
 import os
 import re
 import glob
-import copy
 import logging
 import datetime
 import itertools
@@ -269,13 +268,6 @@ class Product:
         else:
             self.num_threads = int(num_threads)
 
-        # set variables to check for tropo extract logic
-        self.tropo_models = tropo_models
-        self.tropo_extract = False
-        if layers is not None:
-            if 'troposphere' in layers:
-                self.tropo_extract = True
-
         # Determine if file input is single file, a list, or wildcard
         # If list of files
         if len([str(val) for val in filearg.split(',')]) > 1:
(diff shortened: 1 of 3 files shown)


sssangha:
@alexfore , I tried to simplify the logic for this dummy-proofing the best I can in this first draft, so please advise if there is any way we should streamline or improve on this.

Also, can you please confirm whether the imports for the [product.py ](https://github.com/aria-tools/ARIA-tools/blob/3cd16d6086e5188dcb5f42cc6fa18d1780eccfb3/tools/ARIAtools/product.py)script are organized appropriately? Thanks

alexfore (approved):

alexfore:
I think my only comment is that the copy does not seem needed as VALID_TROPO_MODELS is not being modified but fine as is

tools/ARIAtools/product.py
```diff
@@ -714,6 +722,49 @@ def __mappingVersion__(self, fname, version):
                     if '/science/grids/corrections/external/troposphere/' in i:
                         model_name.append(i.split('/')[-3])
 
+                # exit if user does not specify a valid tropo model name
+                if self.tropo_models is not None:
+                    if self.tropo_models.lower() != 'all':
```
alexfore: Sorry, I found an issue with this line when running the regression tests. I found that `self.tropo_models` may be a list, in which case the `self.tropo_models.lower()` raises `AttributeError`. Wrapping this block in
```
                    with contextlib.suppress(AttributeError):
```
avoids that issue, or a similar solution will work

alexfore (changes requested):

tools/ARIAtools/product.py
sssangha: Hello @alexfore , I went ahead and adjusted the logic. Before my tweaks, the user input for `tropo_model` was tested for every file in a loop. This is only necessary to check once, and for only Sentinel-1 products (it is not currently relevant for NISAR products). Please let me know what you think. Thanks

alexfore: OK, it passes the tests now

alexfore (approved):

PR #432, opened 2024-07-29
Add standard layers to ariaTSsetup.py

alexfore:
Here we add the option to specify 'standard' to extract a standard set of layers, which is set to a constant `ARIA_STANDARD_LAYERS` defined in `ARIAtools/constants.py`. If the string specified in the layers command line argument is standard we replace it with those standard layers.

### DIFF
--- a/tools/ARIAtools/constants.py
+++ b/tools/ARIAtools/constants.py
@@ -34,8 +34,8 @@ ARIA_LAYERS += ARIA_INTERNAL_CORRECTIONS
 ARIA_LAYERS += ARIA_TROPO_MODELS
 
 ARIA_STANDARD_LAYERS = [
-    'unwrappedPhase', 'coherence', 'incidenceAngle', 'lookAngle',
-    'azimuthAngle', 'bPerpendicular']
+    'unwrappedPhase', 'coherence', 'incidenceAngle', 'azimuthAngle',
+    'bPerpendicular']
 
 ARIA_STACK_DEFAULTS = ['unwrappedPhase',
                        'coherence',
--- a/tools/ARIAtools/constants.py
+++ b/tools/ARIAtools/constants.py
@@ -34,7 +34,8 @@ ARIA_LAYERS += ARIA_INTERNAL_CORRECTIONS
 ARIA_LAYERS += ARIA_TROPO_MODELS
 
 ARIA_STANDARD_LAYERS = [
-    'unwrappedPhase', 'coherence', 'incidenceAngle', 'azimuthAngle']
+    'unwrappedPhase', 'coherence', 'incidenceAngle', 'lookAngle',
+    'azimuthAngle', 'bPerpendicular']
 
 ARIA_STACK_DEFAULTS = ['unwrappedPhase',
                        'coherence',
--- a/tools/ARIAtools/constants.py
+++ b/tools/ARIAtools/constants.py
@@ -33,6 +33,9 @@ ARIA_LAYERS += ARIA_EXTERNAL_CORRECTIONS
 ARIA_LAYERS += ARIA_INTERNAL_CORRECTIONS
 ARIA_LAYERS += ARIA_TROPO_MODELS
 
+ARIA_STANDARD_LAYERS = [
+    'unwrappedPhase', 'coherence', 'incidenceAngle', 'azimuthAngle']
+
 ARIA_STACK_DEFAULTS = ['unwrappedPhase',
                        'coherence',
                        'connectedComponents',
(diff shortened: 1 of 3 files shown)


sssangha (changes requested):
This is a good idea towards making TS script more intuitive. Thanks @alexfore 

However, the layers to be extracted for the minimum working example in TS mode should reflect what is captured [here](https://github.com/aria-tools/ARIA-tools/blob/f937ed0986957187d1a2feaeed11fcddeb98caf9/tools/ARIAtools/util/vrt.py#L477).

alexfore:
@sssangha I'm not clear on what the requested resolution is. Should `TS_LAYERS_DUP` be changed to reflect what I put into `ARIA_STANDARD_LAYERS`, vice-versa, or something else?

sssangha:
> @sssangha I'm not clear on what the requested resolution is. Should `TS_LAYERS_DUP` be changed to reflect what I put into `ARIA_STANDARD_LAYERS`, vice-versa, or something else?

`ARIA_STANDARD_LAYERS` should be updated to reflect `TS_LAYERS_DUP`, since the latter captures the prerequisite standard layers for TS mode.

alexfore:
Ok, so I propose moving `TS_LAYERS_DUP` to the constants module and using it for the layers extracted for `-l standard` command line args.  Can you tell me what `DUP` means in the variable name in this case?

sssangha:
> Ok, so I propose moving `TS_LAYERS_DUP` to the constants module and using it for the layers extracted for `-l standard` command line args. Can you tell me what `DUP` means in the variable name in this case?

In retrospect, I'm honestly not sure what `DUP` meant in this context. My guess is "duplicate", but that doesn't make much sense to me.

We could remove `TS_LAYERS_DUP` from there, but I think a variable name `ARIA_STANDARD_TS_LAYERS` would be most appropriate as it hopefully more clear to users that those are the standard layers for TS mode.

Also if a user doesn't specify any argument for `-l` when running `ariaTSSsetup.py`, I'm thinking `standard` should be the default argument. What do you think?

alexfore:
@sssangha I made those changes in my last two commits

mgovorcin:
@sssangha @alexfore agree with the mentioned, I also prefer keeping `ARIA_STANDARD_LAYERS` as constant and removing `TS_LAYERS_DEDUP`. 

Agree that `standard` should be default argument for `ariaTSSetup.py`, `-l`, 

@sssangha I do not see the need to keep 'bPerpendicular' as we are going to get the mean bperp directly from GUNW using tsSetup.py. Users can use ariaExtract.py to extract bPerpendicular if the decide to. Optionally we add, if users specify `-l standard bPerpendicular` to use ariaExtract.py to extract bPerpenciular, if not specified default to get it directly from GUNWs.

Also, do not see the need for 'lookAngle' as that layer is not used later on, but we can keep it as part of standard layers to extract if you prefer. Tagging @dbekaert

sssangha:
@mgovorcin , agreed but the change you mentioned regarding bPerp seems to be captured in another [branch](https://github.com/aria-tools/ARIA-tools/tree/remove_bPerp_from_tssetup) made by Alex. Perhaps it would be more appropriate to follow-up with this change through that branch to by default extract and pass the mean bPerp. Otherwise, by discarding the bPerp from the expected layers in this PR it would lead to issues with MintPy as is (i.e. wouldn't it be expected in the stack.vrt?).

Makes sense?

And yes, `lookAngle` appears to be a vestige that we can discard from the standard layers. I thought it was passed before through `prep_aria.py` on the MintPy end, but it doesn't appear to be the case anymore.

mgovorcin:
@alexfore  looks good to me, thanks for the PR. I added a comment regarding bPerp, that can be addressed in other PR
@sssangha  make sense to me.

alexfore:
OK, so I should remove 'lookAngle' from `ARIA_STANDARD_LAYERS` in this branch, then after this branch is merged into dev, merge dev into my other branch, and in that branch remove `bPerpendicular`, correct?

mgovorcin (approved):

sssangha:
> OK, so I should remove 'lookAngle' from `ARIA_STANDARD_LAYERS` in this branch, then after this branch is merged into dev, merge dev into my other branch, and in that branch remove `bPerpendicular`, correct?

Yes, thanks!

mgovorcin:
> OK, so I should remove 'lookAngle' from `ARIA_STANDARD_LAYERS` in this branch, then after this branch is merged into dev, merge dev into my other branch, and in that branch remove `bPerpendicular`, correct?

Correct, this would be good way to go . Use `remove_bPerp_from_tssetup` branch to push this change.

sssangha (approved):
LGTM!

PR #433, opened 2024-07-29
Update README.md

alexfore:
Here we add an installation note about how to suppress the citation messages from GNU parallel


sssangha (approved):
LGTM! Thanks

PR #435, opened 2024-08-01
Remove b perp from tssetup

alexfore:
Here we remove the extraction by default of bPerpendicular instead extracting the average over frames into a JSON file.  If the user explicitly requests bPerpendicular we will extract it.

For a case with 54 GUNWs, I found a runtime of 29:06.43 with the full bPerp layer extraction and 8:37.83 with just the mean JSON extraction.  The JSON extraction of bPerp is done in parallel as well with dask.delayed and threads.


sssangha (approved):
Tested it out and MintPy handles the inputs as expected. Changes LGTM!

PR #437, opened 2024-09-03
Rz extract dedup - hardcode lonlat spacing

rzinke:
Hard coded lon lat pixel resolution values to 0.000833334 degrees. This was done in a single place: lines 487-488 in the `extractProduct.py` module, `mergedproductbbox` function. To the best of my understanding, the parameters coded here should propagate throughout the `ariaTSsetup.py` bin file, in which mergedproductbbox returns the pixels resolution, and that returned value is passed to calls to the `export_products` function.

This change is made as part of a larger ARIA-tools forward processing deduplication effort.


dbekaert:
@rzinke with the on-demand capability we might actually have set variations in resolution for the products.
We will want to be able to support costom on-demand, our disaster pipeline might have different resolutions too.
I think if we can pass it through and be strategic on the where it finds this information then we keep future flexibility.

PR #439, opened 2024-09-05
Check if mask argument passed to PlotClass

rzinke:
Brief conditional statement to check if mask is passed.


sssangha (approved):
LGTM! Made one minor adjustment so that the assignment of the mask variable abides by pep8 convention

PR #448, opened 2024-10-16
Debug to extract ionosphere for single product

rzinke:
export_ionosphere was unable to extract the ionosphere layer for a single product. Single-product ionosphere extraction requires extraction to disk for cropping, etc. Physical extraction now in-place.


sssangha (approved):
Tested out and looks good on my end!

PR #451, opened 2024-10-21
Rz ts skip nonstack

rzinke:
Change Exception to LOGGER.warning when non-stack product encountered in ariaTSsetup.py


sssangha (approved):
LGTM

PR #453, opened 2024-10-28
set log level to info to print steps while running

ehavazli:
This change is in response to the request for `add missing printout of running steps`


alexfore (approved):
I have no opinion on the default log level

Issue #443, opened 2024-09-15
[BUG] ariaDownload not working with asf_search 8.0.1


EJFielding:
**Describe the bug**

I installed `aria-tools` and `asf_search` in a conda environment. When I try to run `ariaDownload.py`, I get an error from the asf_search call.

**To Reproduce**
1. Install `aria-tools` 1.2.1 and asf_search v 8.0.1 
2. Command used `ariaDownload.py -b '37.0 38.1 -122.6 -121.25' --track 35 --num_threads 8`
3. Error Output:
```
Traceback (most recent call last):
  File "/u/vento-r0/ericf/packages/miniconda3/envs/ARIA-tools/bin/ariaDownload.py", line 393, in <module>
    main()
  File "/u/vento-r0/ericf/packages/miniconda3/envs/ARIA-tools/bin/ariaDownload.py", line 389, in main
    Downloader(args)()
  File "/u/vento-r0/ericf/packages/miniconda3/envs/ARIA-tools/bin/ariaDownload.py", line 229, in __call__
    scenes = self.query_asf()
             ^^^^^^^^^^^^^^^^
  File "/u/vento-r0/ericf/packages/miniconda3/envs/ARIA-tools/bin/ariaDownload.py", line 349, in query_asf
    dataset=asf_search.DATASET.ARIA_S1_GUNW,
            ^^^^^^^^^^^^^^^^^^
AttributeError: module 'asf_search' has no attribute 'DATASET'
```

**Screenshots**

**Desktop (please complete the following information):**
 - ARIA-tools git tag: [e.g. git show --summary] `aria-tools                1.2.1           py311h4a3439e_0    conda-forge`
 - OS Linux
asf_search                8.0.1              pyhd8ed1ab_0    conda-forge
asf_search-base           8.0.1              pyhd8ed1ab_0    conda-forge

**Additional context**

mgovorcin:
resolved in $455

mgovorcin:
resolved in #455

PR #455, opened 2024-11-05
Update download call to be compatible with asf_search update

sssangha:
Addresses issue #443 

Now `asf_search.constants` is leveraged, as opposed to `asf_search.DATASET`, to be compatible with the latest `asf_search` increment `v8.0.1`. This is backwards compatible with older versions.

@rzinke , @ehavazli, can you please confirm this PR works on your end? Thanks


rzinke:
Worked for me. ```ariaDownload.py -t 64 -b "34.5 35.7 -116.5 -114.2" -s 20180101 -e 20190101```

PR #454, opened 2024-10-30
Rz logging dedup2

rzinke:
Latest updates to prevent or minimize reprocessing for ariaTSsetup and ariaExtract.
Logging included.

PR #456, opened 2024-11-08
Added runlog

rzinke:
I added "run logging" capability to the current dev branch, without any further application of the feature for dedup, etc. This is the first, and necessary step toward further deduplication efforts.


alexfore:
I think this looks pretty good, here are some specific comments:

* pythonic way of checking for not None is `if self.runlog is not None` vs if self.runlog
* The logger for a module should always be `logging.getLogger(__name__)`, unless that module is an executable script where it should be named something specific (the above will name it `__main__` which isn't that helpful).
* The logger configuration (i.e. call to logging.basicConfig) should only be done just inside the main() method of an executable script (just after parsing input args telling us which log level to use), never anywhere else.  Whatever executable was called which ends up instantiating this class will have/should have already configured the logger.

For my lazy formatting comment before, this isn't lazy
```
            LOGGER.warning(
                'Selected %s not supported in tsSetup' % layer + msg)
```
But this is
```
            LOGGER.warning(
                'Selected %s%s not supported in tsSetup', layer, msg)
```
The key difference being that the item to be formatted is a _function argument_ in the latter call to the warning method (it will only be formatted in the warning method if the logger message will be printed) while in the former it is formatted before the string can be passed to the warning method (it will always be formatted as it must be before the warning method can be called). Again I don't think this is really that important in all cases, but in some cases it can be very important. Specifically in the case of lots of debug log messages being formatted but never printed slowing down execution when run at a info or warning log level.

rzinke:
Thanks for the detailed comments and explanations, @alexfore. I believe I have implemented them and updated the PR.

tools/ARIAtools/extractProduct.py
```diff
@@ -344,13 +344,49 @@ def merged_productbbox(
                             f'minimum threshold area '
                             f'{minimumOverlap}km\u00b2')
 
+    # Check if product bounding box exists from previous run
```
mgovorcin: these changes are the same as in #454

mgovorcin: same applies for L319, L386, L511-552, L606-632

mgovorcin:
@rzinke this PR has a significant overlap with #454. Suggest to isolate runlog.py and its initialization within ariaExtract and ariaTSSetup, and keep forward processing addition in an another PR.

Can you please add an example of config.yaml to this PR, so we have its output here as a reference

Please also clear the branch conflicts

rzinke:
Here are some example config files: config.yaml, files.yaml, and extracted_files.yaml, zipped up because GitHub won't allow posting of YAML files here.
[ConfigExamples.zip](https://github.com/user-attachments/files/17707964/ConfigExamples.zip)

The codes to generate them are:
```
ariaDownload.py -t 71 -b '37.5 39.5 -118.5 -114.5' -s 20200515 -e 20200601

ariaExtract.py -f "products/*.nc" -w test_extract_log -l 'unwrappedPhase' --log-level debug
ariaExtract.py -f "products/*.nc" -w test_extract_log -l 'unwrappedPhase' --log-level debug
```

Note that `ariaExtract.py` is run twice, to create two date entries.

sssangha:
Heads up I tested and fixed a minor logging conflict for `ariaTSsetup` in my last commit. The list of available layers and the unavailable layer name were getting mixed up before.

mgovorcin:
> Here are some example config files: config.yaml, files.yaml, and extracted_files.yaml, zipped up because GitHub won't allow posting of YAML files here. [ConfigExamples.zip](https://github.com/user-attachments/files/17707964/ConfigExamples.zip)
> 
> The codes to generate them are:
> 
> ```
> ariaDownload.py -t 71 -b '37.5 39.5 -118.5 -114.5' -s 20200515 -e 20200601
> 
> ariaExtract.py -f "products/*.nc" -w test_extract_log -l 'unwrappedPhase' --log-level debug
> ariaExtract.py -f "products/*.nc" -w test_extract_log -l 'unwrappedPhase' --log-level debug
> ```
> 
> Note that `ariaExtract.py` is run twice, to create two date entries.

Thanks @rzinke. I checked the config files, looks great in yaml, can you please:
 - use absolute path for work_dir in config.yaml
 - extracted_file.yaml, include somewhere run_time, wither in the filename or add another section in the extracted_files.yaml
        e.g  run_time: 0241111-115835
               - unwrapped_phase:
                  - path1
                  - path2
                  nb_extracted: 3 
               - connected_components:
                  nb_extracted: 0
               - coherence:
                  nb_extracted: 0
               - tropo:
               - set:
               - iono:
   - files.yaml looks good, would do the same with including run_time as in extracted.yaml
 
Think the above might help with the bookkeeping

rzinke:
[YAML_files_af1a8d5..zip](https://github.com/user-attachments/files/17724002/YAML_files_af1a8d5.zip)
Here are the log files based on outputs from af1a8d5

mgovorcin:
> [YAML_files_af1a8d5..zip](https://github.com/user-attachments/files/17724002/YAML_files_af1a8d5.zip) Here are the log files based on outputs from [af1a8d5](https://github.com/aria-tools/ARIA-tools/commit/af1a8d5011366fc6c6cfe0906eb8ad2996a5c328)

thanks @rzinke , is it possible to structure extracted_products.yaml and file.yaml like this: 
```
runs:
  - run_time: 20241112-153230
    coherence:
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/coherence/20200523_20200517
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/coherence/20200529_20200517
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/coherence/20200529_20200523
    connectedComponents:
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/connectedComponents/20200523_20200517
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/connectedComponents/20200529_20200517
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/connectedComponents/20200529_20200523
    ionosphere: []
    solidEarthTide: []
    unwrappedPhase:
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/unwrappedPhase/20200523_20200517
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/unwrappedPhase/20200529_20200517
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/unwrappedPhase/20200529_20200523

  - run_time: 20241113-101500
    coherence:
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/coherence/20200601_20200524
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/coherence/20200607_20200524
    connectedComponents:
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/connectedComponents/20200601_20200524
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/connectedComponents/20200607_20200524
    ionosphere: []
    solidEarthTide: []
    unwrappedPhase:
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/unwrappedPhase/20200601_20200524
      - /Users/rzinke/Documents/ARIAdev/FwdProc-Dedup/SmTest/test_extract_log/unwrappedPhase/20200607_20200524```
`

rzinke:
Updated run logs attached. 
[RunLogs_e8a6951.zip](https://github.com/user-attachments/files/17735777/RunLogs_e8a6951.zip)

mgovorcin (approved):
LGTM

PR #459, opened 2024-11-27
Remove vestigial GACOS correction functionality

sssangha:
Clean removal of vestigial GACOS correction routine, which is already actively supported through [MintPy](https://github.com/insarlab/MintPy/blob/main/src/mintpy/tropo_gacos.py)


mgovorcin (approved):
Looks good to me, thanks @sssangha

PR #465, opened 2025-01-29
Clarify tropo extraction

ehavazli:
- Default value for `-tm, --tropo_models` is changed to `all` so all existing tropospheric models can be extracted (currently only one model is included in ARIA products)
- Added a switch to if no tropospheric layer is requested for extraction.

This method allows us to preserve existing functionality and leaves room future changes to products.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ export VSI_CACHE=YES
 
 The ARIA-tools scripts are highly modulized in Python and therefore allows for building your own processing workflow. Below, we show how to call some of the functionality. For detailed documentation, examples, and Jupyter notebooks see the [ARIA-tools-docs repository](https://github.com/aria-tools/ARIA-tools-docs). We welcome the community to contribute other examples on how to leverage the ARIA-tools (see [here](https://github.com/aria-tools/ARIA-tools/blob/master/CONTRIBUTING.md) for instructions).
 
-* NOTE, currently ARIA-tools does not support deduplication (we are working to add it soon though), so for each commandline input please make sure to specify a fresh output directory with the `--workdir` option.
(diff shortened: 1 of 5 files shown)


sssangha:
Hello @ehavazli , looks good!

I have a few requests:
- In the constants [script](https://github.com/aria-tools/ARIA-tools/blob/dev/tools/ARIAtools/constants.py#L24-L27), please remove `GMAO` and comment out `HRES`, noting this model is not currently available in the GUNW products.
- Translate these changes from `ariaTSsetup` to `ariaExtract` as well
- Add a line in the README that reflects the parser message about how you need to explicitly specify extraction of correction layers in `ariaTSSetup`. We may also acknowledge how e.g. "Currently we support processing of tropospheric estimates derived from the `HRRR` weather model for v3.0.1 products spanning the continental United States, and adjacent regions of Canada and Mexico."

ehavazli:
> Hello @ehavazli , looks good!
> 
> I have a few requests:
> 
>     * In the constants [script](https://github.com/aria-tools/ARIA-tools/blob/dev/tools/ARIAtools/constants.py#L24-L27), please remove `GMAO` and comment out `HRES`, noting this model is not currently available in the GUNW products.
> 
>     * Translate these changes from `ariaTSsetup` to `ariaExtract` as well
> 
>     * Add a line in the README that reflects the parser message about how you need to explicitly specify extraction of correction layers in `ariaTSSetup`. We may also acknowledge how e.g. "Currently we support processing of tropospheric estimates derived from the `HRRR` weather model for v3.0.1 products spanning the continental United States, and adjacent regions of Canada and Mexico."

Hi @sssangha, please review the new changes.

sssangha (approved):
LGTM!

PR #467, opened 2025-02-11
Add support for NISAR sample product v4.0.4

sssangha:
You could test out extraction of supported layers with the following command after downloading the sample product:
`ariaExtract.py -f "products/*h5" -l 'unwrappedPhase,troposphereTotal,ionosphere' -w NISAR_v404 -of ISCE -d Download -m Download`

Note these proposed changes are not backwards compatible with the previous sample products.

### DIFF
--- a/tools/ARIAtools/product.py
+++ b/tools/ARIAtools/product.py
@@ -496,10 +496,10 @@ class Product:
             version = basename.split('_')[-1][:-3]
             version = '.'.join(version)
             nc_version_check = [version]
-            if not basename.endswith('_J_001.h5'):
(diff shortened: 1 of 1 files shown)


ehavazli (changes requested):
We should preserve backwards compatibility for now.

sssangha:
> We should preserve backwards compatibility for now.

Talked with David and he questioned the need for this given there weren't many products for the past version and people wouldn't be looked at them anymore.

Let's discuss this again though.

ehavazli (approved):
Looks good

PR #469, opened 2025-02-26
Extend ionosphere to cover unwrapped phase completely

ehavazli:
Added the functionality to interpolate ionosphere slightly beyond the unwrapped phase to ensure coverage and non-zero errors in time series inversion.


sssangha (approved):
Tested out with stitching case example, and it worked well for that too!

PR #470, opened 2025-02-27
Implement sanity check to confirm download is done

sssangha:
It was brought to my attention that downloading could sometimes silently fail, such that partially downloaded files persist and raise errors later in the `product.py` script when users run other tools. Thanks for the heads up @gracebato!

To test this out, we can use this example:
`ariaDownload.py --track 29 --bbox "36.3 36.5 25.3 25.5" --version 3_0_1 -nt 10  -o Url --start 20170106 --end 20170204 `


ehavazli (approved):
LGTM!

PR #475, opened 2026-02-20
Modify bperp data reading to ignore NaN values

sssangha:
Updated the data reading process to use np.nanmean to ignore NaN values.


mgovorcin (approved):
Thanks Sim, looks good

PR #476, opened 2026-02-24
Add NISAR binary mask support in raster loading

sssangha:
Enhanced the load function to apply NISAR binary mask before warping the raster. Also added error handling for mask validation.

I created a separate function to handle the logic of the NISAR internal mask and translate it to a simple binary file.

This file is needed to properly mask NISAR unw layers to remove artifacts along the edges.

E.g. see both the concomp and unw layers before and after this bug-fix for a single frame:
Screenshot 2026-02-24 at 1 29 40 AM
Screenshot 2026-02-24 at 1 28 58 AM

If not done, this issue will have implications for stitching.

E.g. see stitching in the unw field before and after implementing this fix:
Screenshot 2026-02-24 at 1 26 51 AM

### DIFF
--- a/tools/ARIAtools/util/seq_stitch.py
+++ b/tools/ARIAtools/util/seq_stitch.py
@@ -95,32 +95,58 @@ def stitch_unwrapped_frames(input_unw_files: List[str],
                 'PATH'].split('"')[1].split('/')[-1]
             LOGGER.info('Frame-2: ', frame2_prods)
 
+        # determine if NISAR GUNW
+        is_nisar_file = False
+        frame1_nisar_msk = None
+        frame2_nisar_msk = None
+        track_fileext = unw_attr_dicts[ix1]['PATH'].split('"')[1]
+        if track_fileext.endswith('.h5'):
+            is_nisar_file = True
+            # Get paths to masks from each respective frame
+            # NOTE: Only load Frame 1 mask if it's the very first iteration.
+            # In subsequent iterations, Frame 1 is the
+            # already-masked `corr_unw`.
+            if i == 0:
+                frame1_nisar_msk = \
+                    ARIAtools.util.stitch.get_binary_nisar_mask_from_path(
+                    unw_attr_dicts[ix1]['PATH']
+                )
+            
+            frame2_nisar_msk = \
+                ARIAtools.util.stitch.get_binary_nisar_mask_from_path(
+                unw_attr_dicts[ix2]['PATH']
+            )
+
         # Get numpy masked arrays
         # Frame1
         frame1_unw_array = ARIAtools.util.stitch.get_GUNW_array(
             filename=unw_attr_dicts[ix1]['PATH'],
             proj=proj,
             xres=xres, yres=yres,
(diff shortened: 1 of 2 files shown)


tools/ARIAtools/util/seq_stitch.py
```diff
@@ -95,32 +95,58 @@ def stitch_unwrapped_frames(input_unw_files: List[str],
                 'PATH'].split('"')[1].split('/')[-1]
             LOGGER.info('Frame-2: ', frame2_prods)
 
+        # determine if NISAR GUNW
+        is_nisar_file = False
+        frame1_nisar_msk = None
```
mgovorcin, line 100: @sssangha  is it always frame1 and frame2? what happens if there are more frame than 2 to stitch

tools/ARIAtools/util/stitch.py
```diff
@@ -468,3 +500,137 @@ def combine_data_to_single(data_list: list,
...
+    return create_binary_nisar_mask(nisar_mask_path)
+
+
+def create_binary_nisar_mask(mask_path: str) -> np.ndarray:
+    """
+    Reads a 3-digit NISAR SAR mask file and decodes it into a
+    sbinary mask in memory.
+    
+    1 = Valid non-water (Land)
+    0 = Water or Invalid (Missing data in reference or secondary image)
```
mgovorcin: is it water, or could be noise around edges as well?

dbekaert:
@sssangha this looks good to me. Can you also verify this for the ionospheric correction. In the end the valid pixels should be consistent between all the layers, otherwise masking should be done to them too.

mgovorcin (approved):
Looks good to me

tools/ARIAtools/util/stitch.py
dbekaert: based on the product spec its the common overlap that is missing + water.

tools/ARIAtools/util/seq_stitch.py
dbekaert: my understanding is that the consistent for  h5 or netcdf; i.e. nisar or sentinel-1 is done early on. @sssangha to confirm.

When you get here; you already have only nisar or sentinel files. Hence confirming here again if its nisar is ok to be done on one image. 

Then stitching is incremental. A-B; AB-C; ABC-D. @sssangha to confirm

sssangha: The program works incrementally in sets of 2 frames, this part was inherited from previous implementation so it should still work for NISAR. Would need to have sanity test to be sure

dbekaert:
@hfattahi brief question related to the water mask that makes up the masking layer. 

Can you confirm the water mask is applied to the wrapped phase prior to unwrapping and thus any pixels water are deemed not useful? Trying to understand the implication in wetlands where tidal land could be mased using a permanent water body mask.

dbekaert:
We should potentially limit currently the masking to be on the edges. Leave the water in unless user asked for it to be masked conform the other ARIA-tools GUNW-S1 option. Then user could still bring own mask or instruct to use NISAR embedded water mask.

PR #480, opened 2026-02-25
add NISAR support to ARIA-download

dbekaert:
add GUNW NISAR support

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 [![Language](https://img.shields.io/badge/python-3.9%2B-blue.svg)](https://www.python.org/)
 [![License](https://img.shields.io/badge/License-Apache_2.0-green.svg)](https://github.com/aria-tools/ARIA-tools/blob/master/LICENSE)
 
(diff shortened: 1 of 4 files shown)


dbekaert:
@sssangha once verified, perhaps we should do another minor release on conda.

sssangha:
I made some minor tweaks to address the following:

When I ran the example `ariaDownload.py --mission NISAR -o count`, I got this error:
```
  File "/u/trappist-r0/ssangha/conda_installation/miniforge/miniforge/ARIA-tools_dev/tools/bin/ariaDownload.py", line 462, in <module>
    main()
  File "/u/trappist-r0/ssangha/conda_installation/miniforge/miniforge/ARIA-tools_dev/tools/bin/ariaDownload.py", line 457, in main
    raise Exception('Must specify either a bbox or track')
```

This was fixed by setting `--bbox` by default to global.


I cleaned up the epilog in the parser to provide some contextual comments behind what each of the example inputs correspond to.


Also, nothing is returned on my end when I specified I used the `-t` option for NISAR. Digging a bit further, I saw that the `asf_search` module returned `"pathNumber": null`. This was my cue that the existing minimum `asf_search` version in the environment was out of date. Once updating to `12.0.1` this issue goes away. I've updated the environment YML accordingly to reflect this.

sssangha (approved):
LGTM, but @dbekaert please look at my comment above and let me know what you think.

If you agree, I'll merge.

dbekaert:
Correct the ASF search needs to be the last version, which has the NISAR support included.

dbekaert:
> This was fixed by setting `--bbox` by default to global.

ok that was intentional, but i think defaulting to global is good approach too and easier

dbekaert:
> the `-t` option for

Ok excellent capture on the version of asf search. also there are only two tracks with gunw curerntly so you would need to be lucky to have picked one.

PR #481, opened 2026-02-27
Projection mismatch for S1 GUNW ionosphere frames + artifact screening for single frame NISAR extraction

sssangha:
#Summary
This PR fixes a bug in ionosphere.py where S1 GUNW frames would fail to stitch properly when requested in a target projection  or specific resolution.

#Details
Previously, S1 GUNW files were loaded directly as native EPSG:4326 arrays using xr.open_dataset(). This caused a severe spatial misalignment, as the arrays were being mapped to bounding boxes (SNWE) that had already been calculated for the warped target projection.

To fix this, the legacy fallback block in stitch_ionosphere_frames has been updated to use get_GUNW_array(..., as_xarray=True). This ensures that standard ARIA files are dynamically warped and reprojected in memory before stitching, perfectly aligning their behavior with the new NISAR pipeline.

Iono stitching before bug-fix:
Screenshot 2026-02-26 at 6 24 13 PM

Iono stitching after bug-fix:
Screenshot 2026-02-26 at 6 23 52 PM

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -1571,6 +1571,7 @@ def export_product_worker(
                 bounds=bounds, clip_json=prods_TOTbbox, output_unw=outFilePhs,
                 output_conn=outFileConnComp,
                 output_format=outputFormatPhys,
+                is_nisar_file=is_nisar_file,
                 range_correction=range_correction, save_fig=False,
                 overwrite=True)
 
@@ -1642,7 +1643,7 @@ def export_product_worker(
 
 def export_products(
         full_product_dict, proj, bbox_file, prods_TOTbbox, layers, arrres,
-        is_nisar_file, rankedResampling=False, demfile=None,
+        iono_filter, is_nisar_file, rankedResampling=False, demfile=None,
         demfile_expanded=None, lat=None, lon=None, maskfile=None, outDir='./',
         outputFormat='VRT', verbose=None, num_threads='2', multilooking=None,
         tropo_total=False, model_names=[], multiproc_method='single',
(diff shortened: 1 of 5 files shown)


sssangha:
In addition, I realized for single frames, the ionosphere layer was being passed without the clean-up that the stitched iono fields were patched through (i.e. fitting of a clean field after masking for conn comp and for NISAR the artifact mask).

Here is how a single frame extraction for NISAR looks before vs after:
Screenshot 2026-02-26 at 7 33 10 PM


Also, in the past we had intentionally avoided masking/cropping the iono layer to the unw cutline because of downstream issues raised by @ehavazli before:
https://github.com/aria-tools/ARIA-tools/pull/469

@ehavazli , could you please confirm if my recollection regarding that piece is correct, or otherwise clarify? Thanks

ehavazli:
@sssangha 
ARIA GUNW ionopshere was falling short at the edges and causing superficial values to bleed into the time series inversion and downstream. The PR linked above solved this issue by interpolating the missing edges with the nearest neighbor.

On the other hand, do we want to extract something that looks different than what's in NISAR data?

dbekaert:
@sssangha so this is like a smooth fitting to the ionospheric phase screen as a means to suppress artifacts?

sssangha:
@ehavazli and @dbekaert good point, yeah we don't want to loose ability to capture and correct e.g. streaks and so.

Thus I decoupled the way the program smooths, filters, and buffers the S1 GUNW iono from NISAR.

But if a user still wants to filter the NISAR iono layer the same way, they may by toggling the ` --iono_filter` option.


Here is how the stitched NISAR iono layer looks with/without the ` --iono_filter` option:
Screenshot 2026-02-27 at 3 37 24 PM

Issue #484, opened 2026-02-27
[BUG] rejection of good NISAR GUNW data


dbekaert:
Currently it appears that we are rejecting valid NISAR GUNW with a warning on old version.
Lets now update this and set this to an actual version 

```
2026-02-27 17:56:05,079 - ARIAtools.product - WARNING - input file NETCDF:"/scratch/dbekaert/Mexico_City/NISAR/GUNWs/T113_079/4000/NISAR_L2_PR_GUNW_005_113_D_079_006_4000_SH_20251118T011330_20251118T011405_20251130T011331_20251130T011405_X05010_F_F_J_001.h5 is an older, unsupported version of the NISAR sample product
```

These are actually two interferograms with a common reference scene but a differnt secondary scene (i.e. they are different pairs). They are both valid. One if uses near real time orbits; the other uses precise orbits.

- NISAR_L2_PR_GUNW_ **005** _113_D_078_ **006** _4000_SH_20251118T011257_20251118T011331_20251130T011257_20251130T011332_X05010_ **F** _F_J_001.h5
- NISAR_L2_PR_GUNW_ **005** _113_D_078_ **007** _4000_SH_20251118T011257_20251118T011331_20251212T011258_20251212T011332_X05010_ **N** _F_J_001.h5

dbekaert:
Diving a little bit more into this:

ARIA-tools currently list criteria as:
https://github.com/aria-tools/ARIA-tools/blob/72d545905b75e5a3978342a30209f72d04cc6b6d/tools/ARIAtools/product.py#L499-L503


the product version is based on the last 001.h5 part. However currently this number will not be rapidly changing. The processing software is what is currently evolving aka "CRID". see here for detailed convention of GUNW

https://nisar-docs.asf.alaska.edu/build/nisar-naming-convent-ffd450101f26e19c7eb8b343edb665ba.png

This is X05010. Currently what we should be supporting is a version 5006 and after.

sssangha:
@dbekaert check out my latest commits + fixes in PR #485

EJFielding:
Possibly a similar problem with NISAR partial frames `_N_P_J_001.h5` #490

dbekaert:
@sssangha this does not solve it. I made the PR in draft mode as it was still filtering it out. I dont see any changes in the PR that was merged that would change that. When running the dev branch it also give me the same rejection of two pairs.

There are two items here. 

[...]

Issue #483, opened 2026-02-27
Better handle warnings when using  NISAR data


dbekaert:
Fixing some nisar warnings:

1) Given the different projections of files we are getting quite some gdal warning printed to the screen.
These are not captured nicely in the logger and overflow the screen:
```
Warning 1: gdalwarp -of VRT just takes into account the first source dataset.\nIf all source datasets are in the same projection, try making a mosaic of\nthem with gdalbuildvrt, and use the resulting VRT file as the input of\ngdalwarp -of VRT.

```

2) the tropo model.
NISAR has an embedded model in it but it not recognized by ARIA-tools hence provides warning:
```
2026-02-27 17:58:24,483 - ARIAtools.util.vrt - WARNING - ERA5 tropo model not found in product
```

sssangha:
@dbekaert I figured out a workaround and included a fix for item # 2 in the latest conda release (PR #486 ), but the former I still have to debug

sssangha:
Ok it looks like I also figured out how to address the 1st bullet in this PR #492

dbekaert:
@sssangha i tested this and confirm this is working

dbekaert:
https://github.com/aria-tools/ARIA-tools/pull/492

Issue #491, opened 2026-03-02
[BUG] failed extraction coh for S1


dbekaert:
**Describe the bug**
aria stack prep fails for coherence extraction of Sentinel-1 data. The code is expecting now ENVI files as output, while only vrt files are crearted. Please check backward compatibility as this migth have been introduced as a result of adding nisar support. 

Work-around is currently if users for `-of envi`.

**To Reproduce**
very small dataset:

1. `ariaDownload.py -t 78 --mission S1 -b '18.99 19.65 -99.4 -98.75'  -nt 5  -s 20251020 -e 20260120 -w GUNWs`
 2. `ariaTSsetup.py -f 'GUNWs/*.nc' -b '18.99 19.65 -99.4 -98.75' -nt 10  -w aria_tools`

this is the latest dev version.

sssangha:
I tested this out and confirmed I patched this in my latest PR:
https://github.com/aria-tools/ARIA-tools/pull/492

dbekaert:
I tested it and confirm this is working

dbekaert:
https://github.com/aria-tools/ARIA-tools/pull/492

PR #492, opened 2026-03-03
Silence warnings and fix amplitude overlap bug

sssangha:
Fix benign warnings captured in #483

Fix benign stitching warning triggered when multiproc_method='threads` method used for unwrap stitching and VRT for first frame is processed in the loop:
```
/u/trappist-r0/ssangha/conda_installation/miniforge/miniforge/ARIA-tools_dev/tools/ARIAtools/util/stitch.py:485: RuntimeWarning: Mean of empty slice
  comb_data = np.nanmean(comb_data, axis=0)
```

Also fix overlooked amplitude output issue associated with NISAR GUNW where the stitched overlap region is nan (left) because 0s/nans in one frame mask an overlapping frame. This logic has been corrected avoid improperly masking out pixels in the overlap region (right).
Screenshot 2026-03-02 at 5 24 49 PM

### DIFF
--- a/tools/ARIAtools/extractProduct.py
+++ b/tools/ARIAtools/extractProduct.py
@@ -1747,7 +1747,7 @@ def export_products(
         prev_maskfile = log_data['maskfilename'] if 'maskfilename' \
             in log_data.keys() else None
 
-        if maskfile != prev_maskfile:
+        if maskfile != prev_maskfile and prev_maskfile is not None:
             update_mode = 'full_extract'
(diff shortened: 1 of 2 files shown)


dbekaert:
comfirm this fixes
#484 
#491 
#483

dbekaert:
@sssangha I also confirm the amplitude to be working as described above.

dbekaert (approved):
looking all good to me

PR #498, opened 2026-03-09
updates to get ariaplot working for NISAR and Sentinel-1.

dbekaert:
This includes the fixes for ariaPlot for Sentinel-1 and also extends it to NISAR GUNW.
I tested this for:
-  -croptouion; single and multiple frames and the test pass locally.
- ariaExtract and ariaTSsetup to avoid intriducing bugs and passes for various cases.

Did not test the masking options.

# S1

image
image
image

# NISAR
image
image

image


dbekaert:
needs update to README that ariaplot support is enabled also for NISAR

sssangha:
@dbekaert I fixed an issue with duplicate '.eps' extensions for one of the bperp plots.

E.g. I was getting files like this:
```
bperp_histogram.eps.eps
bperp_histogram.eps.png
```

Also I tested it out with/without using a mask.

All works well! Thanks

sssangha (approved):
LGTM

PR #510, opened 2026-03-20
allowing subsetting for 3D grid 

dbekaert:
This PR is made to allows subsetting for 3D  grid. this should make the 2D layer extraction from 3D grids a lot faster. especially under virtual processing scenario. For physical the gains are expected to be less.

PR #516, opened 2026-03-25
Fix cached metadata handling in mappingVersion

ehavazli:
## Summary
- avoid UnboundLocalError in `Product.__mappingVersion__()` for v1c products when metadata is served from cache
- reuse cached `subdatasets` as the source of available correction layers and only fall back to `gdal.Info()` when needed
- keep tropo model discovery and additional layer filtering aligned for both cache-hit and fallback paths


dbekaert:
ported in the existing PR #510

PR #518, opened 2026-04-01
Add NISAR ephemeral archive collection support for search

dbekaert:
Include private ephemeral archive collection (C4052499921-ASF) alongside the public NISAR GUNW collection for NISAR searches. Authenticate CMR queries via ASFSession using netrc or explicit credentials so the private collection metadata is visible.

@sssangha not sure if you have set up the access in the past, i believe so; but in case you have not done so please reach out so we can get it sorted for future ST work. tagging


dbekaert:
This works as expected and evident from the baseline plot now havign more data included. 

image

dbekaert:
This has been tested and confirmed for S3 and local set up

PR #527, opened 2026-06-23
Refactor create_binary_nisar_mask to be less aggressive

sssangha:
Updated the create_binary_nisar_mask function to clarify the conditions for valid and invalid data, and modified the binary mask creation logic to consider both reference and secondary swaths.

### DIFF
--- a/tools/ARIAtools/util/stitch.py
+++ b/tools/ARIAtools/util/stitch.py
@@ -552,20 +552,12 @@ def get_binary_nisar_mask_from_path(gdal_path):
 
 def create_binary_nisar_mask(mask_path: str) -> np.ndarray:
     """
-    Reads a 3-digit NISAR SAR mask file and decodes it into a
-    binary mask in memory.
+    Reads a 3-digit NISAR SAR mask file and decodes it into a binary
+    mask in memory.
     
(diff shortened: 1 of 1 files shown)


sssangha:
* **The Issue:** The `create_binary_nisar_mask` function in `util/stitch.py` was incorrectly masking out broad swaths of valid data along the azimuth. This occurred because modulo-10 arithmetic was being applied directly to the raw mask values. If NISAR packed higher-order flags into the bits above the base 3-digit code (Bit 8+), it corrupted the modulo calculation and falsely flagged pixels as invalid. 
* **The Fix:** * Applied a bitwise AND mask (`& 0xFF`) prior to digit extraction to safely strip out any higher-order bits.
    * Updated the validation logic to verify that *both* the reference and secondary subswaths contain valid data (`> 0`), bringing the base rigor in line with MintPy's handling. 
    * The intentional pass-through of the water mask (hundreds digit) was preserved to maintain user control over water bodies.
* **Impact:** Eliminates arbitrary data loss along the azimuth and ensures consistent, reliable mask generation.


IFGs before vs after this patch:
OG_T062_x05016_IFGs
T062_x05016_IFGs

